PR #7, opened 2019-02-12
G3 config updates and cleanup

philboeselager:


hornmo (commented):
Looks great, though I don't understand why we would want to use the log4j logger in ApiController.

philboeselager:
> Looks great, though I don't understand why we would want to use the log4j logger in ApiController.

Right. It's a relic of my temporary local logging. Do you think it's necessary to delete it?

hornmo:
I think best would be to just replace it with the current `@Slf4j`.

hornmo (approved):

PR #9, opened 2019-02-19
Integration job

hornmo:


server/gokbg3/grails-app/controllers/org/gokb/WorkflowController.groovy
```diff
@@ -1729,19 +1732,48 @@ class WorkflowController {
...
+          def old_from_combos = Combo.executeQuery("from Combo where fromComponent = ?", [otd])
+          def old_to_combos = Combo.executeQuery("from Combo where toComponent = ?", [otd])
+
+          old_from_combos.each { oc ->
+            def ctype = oc.type
+            def other_end = oc.toComponent
+            def existing_new = Combo.executeQuery("from Combo where type = ? and fromComponent = ? and toComponent = ?",[ctype, neworg, oc])
+
+            if (existing_new?.size() == 0) {
+              ofc.fromComponent = neworg
```
philboeselager: should be `oc.fromComponent = neworg` ?

philboeselager (approved):
Looks fine to me, just 1 comment, please see below.

If possible / suitable, smaller size of commits would be great, with a more concrete commit message. That would increase the intelligibility. (With respect to https://github.com/openlibraryenvironment/gokb/pull/9/commits/3ea67312566003e2a74a1a4bcfaf160408c21151.)

server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
```diff
@@ -756,235 +756,278 @@ class IntegrationController {
   @Secured(['ROLE_API', 'IS_AUTHENTICATED_FULLY'])
```
philboeselager, line 756: We should divide this huge 250 lines method into more intelligible and dealable submethods soon.

server/gokbg3/grails-app/controllers/org/gokb/WorkflowController.groovy
hornmo: You are right, good catch!

server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
hornmo: That is a good idea.

PR #28, opened 2019-03-29
Fix issues with templates while creating new components

hornmo:
- Improve AjaxSupport error message handling (resolve args, use fallback) & reduce debug messages
- Mark display templates of components that should not be created on their own (IDs, TIPPs etc.)
- Improve taglib handling for Refdata & Combos (required fields)
- Disable tabs with combo information while creating components
- Harmonize templates for different title types
- Move title publisher tab to separate template
- Remove alternate name tab from TIPP display template


philboeselager (approved):
I'm not the world's biggest fan of keeping out-commented code instead of just deleting it - this is what the versioning system is for. But for logging messages, it clarifies when reading the code. So yes, +1 .

PR #29, opened 2019-04-05
Rearranged the mapping of TIPP Containers to db fields so that covera…

VolkerHBZ:
#96 coverageDepth and coverageNotes get persisted correctly


hornmo (changes requested):
Since coverageDepth is a RefdataValue, we first need to get the correct object reference here via a RefdataCategory lookup.

VolkerHBZ:
The Refdata Values get initialized in line 390-392 of BootStrap.groovy. What more needs to be done to create the correct object references that you're referring to?

hornmo:
The `tipp_values` map is used to supply the initial values of a new TIPP object. Since `the_kbart` contains only Strings at this point, the incoming value of coverageDepth has to be converted to the correct RefdataValue first. 

`RefdataCategory.lookup('TitleInstancePackagePlatform.CoverageDepth', the_kbart.coverageDepth)` should do the trick. We can leave out the null check, since for empty/illegal values this will then correctly be set to the default value in ensureDefaults.

VolkerHBZ:
The method TitleInstancePackagePlatform.upsertDTO provides exactly this transformation in line 391 and the database fields are set to the according (numerical) values. If there is anything missing, I don't see what this could be.

hornmo:
You are correct that if we were using `upsertDTO` here this would work, but that is not the case:

https://github.com/openlibraryenvironment/gokb/blob/38973926091cc035657936b7b49400fd13c734c1/server/gokbg3/grails-app/services/org/gokb/TSVIngestionService.groovy#L1182

`upsertDTO` right now is used by most actions in [IntegrationController](https://github.com/openlibraryenvironment/gokb/blob/g3_master/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy) for processing incoming JSON objects (like `crossReferencePackage`, which Ygor uses to import package JSON).

server/gokbg3/grails-app/domain/org/gokb/cred/TitleInstancePackagePlatform.groovy
```diff
@@ -417,7 +420,7 @@ class TitleInstancePackagePlatform extends KBComponent {
           }
 
           if (!cs_match) {
-            tipp.addToCoverageStatements('startVolume': c.startVolume, 'startIssue':c.startIssue, 'endVolume': c.endVolume, 'endIssue': c.endIssue, 'embargo':c.embargo, 'coverageNote': c.coverageNote, 'startDate': parsedStart, 'endDate': parsedEnd)
+            tipp.addToCoverageStatements('startVolume': c.startVolume, 'startIssue':c.startIssue, 'endVolume': c.endVolume, 'endIssue': c.endIssue, 'embargo':c.embargo, 'coverageDepth': c.coverageDepth, 'coverageNote': c.coverageNote, 'startDate': parsedStart, 'endDate': parsedEnd)
```
hornmo: This line creates a new TIPPCoverageStatement and adds it to the list of coverageStatements of the existing TIPP. As of right now, TIPPCoverageStatements do not have their own coverageDepth. There is an ongoing debate about wether a TIPP should be able to have statements of different depths or if this would constitute a new TIPP. For now, this change should be removed.

hornmo (changes requested):

server/gokbg3/grails-app/domain/org/gokb/cred/TitleInstancePackagePlatform.groovy
VolkerHBZ: if I understand correctly, there can be several coverage_data blocks per TIPP, as described in https://github.com/k-int/gokb-phase1/wiki/tipp_dto or is that the subject of debate, as you mentioned...?

hornmo: A TIPP can have multiple coverageStatements, yes. The debate is wether coverages of different depths should be included in the same TIPP, or if any TIPP should only ever have one 'global' depth (as it is right now).

hornmo:
Moved changes to new PR https://github.com/openlibraryenvironment/gokb/pull/45

PR #31, opened 2019-04-05
Harmonize SearchController template & json returns

hornmo:
Search results returned as JSON were still using the old Eval function.


philboeselager (approved):
This is index() method is one of these monsters (200+ lines). We should start refactoring code like this.

PR #40, opened 2019-04-16
Use messages for ajaxSupport errors

hornmo:
Replace static error Strings with messages in AjaxSupport


VolkerHBZ (approved):
looks okay...

Issue #46, opened 2019-04-25
Refactoring Coverage in GOKb


darupp:
At the moment, a TIPP can have multiple coverages, but only one coverage note and coverage depth field. In the title import JSON, like coverage multiple instances of coverage note and depth are possible.

How should GOKb handle this?
1) New Coverage --> new TIPP
2) Multiple Coverages with TIPPs and multiple Coverage notes and depth
3) as it is now

darupp:
Will be decided in the SC

Issue #22, opened 2019-03-06
Package export does not process book info


hornmo:
All current package TSV export functions in PackageController & WorkflowController do not output any Monograph fields.

This should probably be fixed in concordance with #8.

croqueGrec09:
To have the reference, I shall reference here own test results:
Tested on dev & QA
With reference to LAS:eR-tickets ERMS-1203 and ERMS-1150

When exporting EBook-packages (KBart and GOKb), no identifiers are being exported.
Tested with:

- dev: https://phaeton-dev.hbz-nrw.de/gokb/public/packageContent/org.gokb.cred.Package%3A812354
- qa: https://phaeton-qa.hbz-nrw.de/gokb/public/packageContent/org.gokb.cred.Package%3A812408

hornmo:
Fixed in #49

PR #49, opened 2019-04-30
Add coverageDepth to coverage statements

hornmo:
Adds coverageDepth property to TIPPCoverageStatements

Further changes:
- Adjust package TSV export functions
  - Create a line for each coverage statement
  - Add book columns
- Adjust TIPP upsert
- Adjust cleanup function
- Adjust tipp template
- Adjust Package & TitleInstance OAI serialization

Addresses #46 and #22

### DIFF
--- a/server/gokbg3/grails-app/controllers/org/gokb/PackagesController.groovy
+++ b/server/gokbg3/grails-app/controllers/org/gokb/PackagesController.groovy
@@ -364,59 +364,64 @@ class PackagesController {
   // @Transactional(readOnly = true)
   def kbart() {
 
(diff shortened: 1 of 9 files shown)


philboeselager (commented):
Huuuuge diffs. Much reformatting. Less content. Not a real problem, but:
We should commit ourselves to an obligatory coding style guide.
Then, at day X, re-format absolutely everything on master in 1 commit.
And use an coding style sheet with auto-format on save in our IDEs from then on.

philboeselager (approved):

hornmo:
Agreed, something we should come back to.

PR #104, opened 2019-10-23
improved validators

hornmo:
- Aligned various name validators
- Improved namespace pattern matching
- Add contentType to public package view
- Use UUID for referencing TIPPs in templates


VolkerHBZ (approved):
looks okay

Issue #196, opened 2020-02-11
Crossreferencing API should accept date formats as YYYY or YYYY-MM


darupp:
The Crossreferencing API should accept date formats as YYYY or YYYY-MM, because these formats apply to the KBART standard: 

```
For journals, this field should include the date of the first issue available online in the format YYYY-MM-DD. Use only those parts of the date datathat apply.For example, if the journal is annual, only YYYY should be used, whereas if the journal is monthly or quarterly, only YYYY-MM should be used. Only in cases where issues of the journal have specified cover dates including the day should YYYY-MM-DD be used. For books, leave this field blank.
```
Source: https://groups.niso.org/apps/group_public/download.php/16900/RP-9-2014_KBART.pdf

hornmo:
Added in v8.5.1

PR #201, opened 2020-02-14
Adjust upsert date handling

hornmo:
* Switch to LocalDateTime for parsing upsert Dates
* Define parsing of date parts in upsert (#196 )


VolkerHBZ (changes requested):
in IntegrationController, the lines 1472-1477 are dead code, as the property publishedFrom is already set to a length of 10. I suggest to treat the property publishedTo by this part of the code, as it is used in the other copies of that code snipplet to treat the end of a time interval.

Speaking of code duplication: isn't it possible to implement this treatment in a separate method and not copy it several times?

VolkerHBZ (approved):

PR #225, opened 2020-04-06
Rest Profile and Users controller

VolkerHBZ:


server/gokbg3/grails-app/services/org/gokb/UserProfileService.groovy
```diff
@@ -58,8 +60,110 @@ class UserProfileService {
       user_to_delete.delete(flush: true, failOnError: true)
 
       log.debug("Done")
+      result.result = "OK"
+      result.errors = user_to_delete.errors
```
hornmo: Since this is a database delete, the object does not exist here anymore. For handling errors in deletion, we would need to catch them before.

server/gokbg3/grails-app/services/org/gokb/UserProfileService.groovy
```diff
@@ -58,8 +60,110 @@ class UserProfileService {
...
+              }
+            } else {
+              skippedCG = true
+            }
+          }
+        } else if (field == "roles") {
+          // scan data
+          Set<Role> newRoles = new HashSet<Role>()
+          val.each { value ->
+            Role newRole = Role.findByAuthority(value.authority)
```
hornmo: May want to check by Role ID (value.id) as well.

hornmo (changes requested):

server/gokbg3/grails-app/services/org/gokb/UserProfileService.groovy
VolkerHBZ: updated according to the suggestion

server/gokbg3/grails-app/services/org/gokb/UserProfileService.groovy
VolkerHBZ: deleted line 64

hornmo (approved):

PR #240, opened 2020-05-15
Rest sources controller

VolkerHBZ:
Bugfixes & Source Controller

### DIFF
--- a/server/gokbg3/grails-app/services/org/gokb/UserProfileService.groovy
+++ b/server/gokbg3/grails-app/services/org/gokb/UserProfileService.groovy
@@ -2,33 +2,16 @@ package org.gokb
 
 import grails.converters.JSON
 import grails.gorm.transactions.Transactional
-import groovy.util.logging.Slf4j
-import org.gokb.cred.ComponentLike
-import org.gokb.cred.CuratoryGroup
-import org.gokb.cred.DSAppliedCriterion
-import org.gokb.cred.Folder
-import org.gokb.cred.History
-import org.gokb.cred.KBComponent
-import org.gokb.cred.Note
-import org.gokb.cred.Package
-import org.gokb.cred.ReviewRequest
-import org.gokb.cred.Role
-import org.gokb.cred.SavedSearch
-import org.gokb.cred.User
-import org.gokb.cred.UserOrganisation
-import org.gokb.cred.UserOrganisationMembership
-import org.gokb.cred.UserRole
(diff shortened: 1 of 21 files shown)


server/gokbg3/grails-app/services/org/gokb/UserProfileService.groovy
```diff
@@ -79,19 +80,18 @@ class UserProfileService {
   def update(User user, def data, User adminUser) {
     def result = [:]
     def error = [:]
-    log.debug("Updating user ${user.id} ..")
-    def immutables = ['id', 'username', 'passwordExpired', 'last_alert_check']
+    log.debug("Updating user ${user.id} ..".toString())
```
hornmo: We generally should not need to convert GStrings manually for logging.

server/gokbg3/grails-app/services/org/gokb/UserProfileService.groovy
```diff
@@ -79,19 +80,18 @@ class UserProfileService {
...
+    def immutables = ['id', 'username', 'password', 'last_alert_check']
     def adminAttributes = ['roles', 'curatoryGroups', 'enabled', 'accountExpired', 'accountLocked', 'passwordExpired', 'last_alert_check']
 
-    def reqBody = data
     if (!adminUser.isAdmin() && user != adminUser) {
       error.message = "$adminUser.username is not allowed to change $user.username"
```
hornmo: A better solution would be to have an array of errors and adding objects like `[message: "$field not found", badData: ["${field}": $value], code: 400]` for each one.

server/gokbg3/src/integration-test/groovy/gokbg3/rest/UsersTestSpec.groovy
```diff
@@ -98,35 +115,37 @@ class UsersTestSpec extends AbstractAuthSpec {
   void "test PUT /rest/users/{id}"() {
     def urlPath = getUrlPath()
     // use the bearerToken to write to /rest/user
+    String bodyText = "{data:{id: $altUser.id," +
```
hornmo: Preferably just define a map here and convert it via `body(bodyText as JSON)` later (like records in IntegrationControllerSpec for example).

hornmo (commented):
We don't really want to make calls from a service to a controller, as in UserDetailsService.update() to UserController.collectUserProp(). Therefore, collectUserProp should probably be moved to the service class.

hornmo (approved):

PR #271, opened 2020-06-26
Package upsert patch

VolkerHBZ:


server/gokbg3/grails-app/services/gokbg3/RestMappingService.groovy
```diff
@@ -320,7 +321,7 @@ class RestMappingService {
           }
         }
       } else {
-        def linkObj = ptype.get(val)
+        def linkObj = val.id?ptype.get(val.id):ptype.get(val)
```
hornmo: This will throw an exception for the integer case, since `val.id` is not defined.

hornmo (changes requested):

server/gokbg3/grails-app/services/gokbg3/RestMappingService.groovy
VolkerHBZ: sorry. I wasn't aware of that change being pushed...

hornmo (approved):

PR #281, opened 2020-07-03
If val contains the key 'id' rather use its value than val itself

VolkerHBZ:
self-explaining one-liner


hornmo:
Resolved in another PR

Issue #273, opened 2020-06-30
Title mismatch Apollo


croqueGrec09:
Given the TitleInstance https://gokb.org/gokb/resource/show/org.gokb.cred.BookInstance%3A6241061#titledetails.
In fact, it should be two title instances given the same name because it occurs in https://gokb.org/gokb/resource/show/org.gokb.cred.Package%3A11751449 and in https://gokb.org/gokb/resource/show/org.gokb.cred.Package%3A6243109. In former, it is a film title and in latter, it is an eBook. We are thus talking about two different title instances what has the consequence that film titles get ISBNs.

hornmo:
No. Items from KBART can only be separated via their `publication_type`, namely `serial` or `monograph`. Those values directly correlate with a specific Item type in the GOKb (BookInstance/JournalInstance). The value of `medium` can then be used to further differentiate the titles, but will not be used for matching of future imports. So unless the titles contain differing global identifiers (so both would need a different `doi` in effect), I am afraid your use case  cannot be reflected properly by the KBART standard and you should therefore refrain from misusing it in this manner.

croqueGrec09:
The matter of this ticket is that two independent records are reflected into one.

hornmo:
This is because only one of the titles has any reliable identifiers, and both are classified as a BookInstance. Titles without proper identifiers cannot be matched properly, and are therefore prone to cause such errors. The safest solution would be to refuse any titles without proper (globally recognized) identifiers, but that would have to be discussed further.

croqueGrec09:
Alright, should thus become subject of discussion.

philboeselager:
Given the fact that countless examples of matching books & films titles exist, we need to make this work.

@hornmo : What is the minimum criteria to make two titles distinguishable in this case?

My assumption: if there is a duplicate title, check the medium type.

- If the medium types are different, check identifiers.
  - If identifiers are different: OK --> create both items independently
  - If there is a duplicate identifiers: --> create a Review Request
- If the medium types are the same, we can not defer a case descibed here. So the procedure must be kept as is.

hornmo:
Yes, that sounds correct. I was going to completely ignore other types of titles in the identifier lookup, but creating a review request could probably be helpful for such cases.

philboeselager:
@hornmo Is this fixed within the two PRs mentioned above? If so, please close this issue.

hornmo:
Those PRs fixed this issue.

Issue #183, opened 2020-02-07
View Title: Delete title


dirk-ecker:

hornmo:
Added in #250 
Endpoint: `/titles/<id>/delete`
Response: 200

dirk-ecker:
Please use the normal REST Verb delete for this no extra service endpoint. 

DELETE /rest/titles/<id>

hornmo:
Fixed in 8.7.0

dirk-ecker:
Deletion of rest/titles/64666 with method DELETE raises a 500:

```
code: 500
exception: {SQLException: {SQLState: "42601", cause: null, errorCode: 0,…},…}
SQLException: {SQLState: "42601", cause: null, errorCode: 0,…}
cause: {SQLState: "42601", cause: null, errorCode: 0,…}
localizedMessage: "Hibernate operation: could not execute statement; bad SQL grammar [n/a]; nested exception is org.postgresql.util.PSQLException: ERROR: syntax error at or near ")"↵  Position: 233"
message: "Hibernate operation: could not execute statement; bad SQL grammar [n/a]; nested exception is org.postgresql.util.PSQLException: ERROR: syntax error at or near ")"↵  Position: 233"
mostSpecificCause: {SQLState: "42601", cause: null, errorCode: 0,…}
rootCause: {SQLState: "42601", cause: null, errorCode: 0,…}
sql: "n/a"
stackTrace: [{className: "org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator",…},…]
suppressed: []
message: "Server Error"
```

dirk-ecker:
It still raises a 500.

`DELETE /rest/titles/7711091`

```
{code: 500, message: "Server Error",…}
code: 500
exception: {cause: null, localizedMessage: "Cannot get property 'cause' on null object",…}
cause: null
localizedMessage: "Cannot get property 'cause' on null object"
message: "Cannot get property 'cause' on null object"
stackTrace: [{className: "gokbg3.ErrorController", fileName: "ErrorController.groovy", lineNumber: 19,…},…]
0: {className: "gokbg3.ErrorController", fileName: "ErrorController.groovy", lineNumber: 19,…}
1: {className: "org.grails.core.DefaultGrailsControllerClass$MethodHandleInvoker",…}
2: {className: "org.grails.core.DefaultGrailsControllerClass",…}
[...]

PR #294, opened 2020-07-15
Curatory reviews

hornmo:
- Remove allocation information from review request templates (#268 )
- Add missing CRUD actions for reviews
- Add action for retrieving reviews of a curatoryGroup (`/curatoryGroups/<id>/reviews`)
- Respect combo status for list of Identifier Reviews
- Add dedicated ReviewRequestService
- Add new domain class `AllocatedReviewGroup` representing n:m links between RR and CuratoryGroups
- Enable usage of stdDesc refdata for Requests
- Removed unneeded method restriction from Secured annotations
- Fix variable usage in TitleController (#183 )
- Add index for Combo status
- Respect title type in upsert (#273 )
- Filter out deleted title in non-sync upsert
- Remove default RR for titles without class one ids


server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewRequestService.groovy
```diff
@@ -0,0 +1,51 @@
...
+          new AllocatedReviewGroup(group: it, review: req).save(flush:true,failOnError:true)
+        }
+      }
+    }
+
+    req
+  }
+
+  def allocateGroups(rr, user) {
+
```
philboeselager: missing on purpose?

hornmo: There are some missing parts here, I will re-request once everything is fixed.

server/gokbg3/grails-app/services/org/gokb/TitleLookupService.groovy
```diff
@@ -556,13 +564,16 @@ class TitleLookupService {
...
 
-              if(added) {
-                ReviewRequest.raise(
-                    the_title,
-                    "'${metadata.title}' added as a variant of '${the_title.name}'.",
-                    "Title was matched via secondary id, but had a different name.",
-                    user, project
-                    )
+              if (added) {
+                def rr = reviewRequestService.raise(
```
philboeselager: There's nothing further done with `rr` - so maybe drop that assignment?
(Same for the following cases.)

philboeselager (approved):
No blockers. Please check missing method content (see first comment) and maybe leave out the obsolete assignments (see second comment).

PR #298, opened 2020-07-29
endpoint /coverage-depth incl. test

### LINKED ISSUE #286: View Add Title: Coverages service missing
For filling the select box in the Add Title dialog I need a service to get all coverages values. The wiki describes these:
 * Abstracts
 * Fulltext
 * Selected Articles

VolkerHBZ:


server/gokbg3/grails-app/controllers/org/gokb/rest/RefdataController.groovy
```diff
@@ -154,4 +154,34 @@ class RefdataController {
...
   }
+
+  @Secured(['IS_AUTHENTICATED_ANONYMOUSLY'])
+  def coverageDepth() {
+    def result = [:]
+    def resultData = []
+    def cat = null
+    def base = grailsApplication.config.serverURL + '/'+namespace
+
+    cat = RefdataCategory.findByLabel("TitleInstancePackagePlatform.CoverageDepth")
```
hornmo: Since coverage info has been moved completely to TIPPCoverageStatement, we should use `TIPPCoverageStatement.CoverageDepth` here.

hornmo (changes requested):

server/gokbg3/grails-app/controllers/org/gokb/rest/RefdataController.groovy
VolkerHBZ: Not sure how to get a list of possible coverage refdata values from that TIPPCoverageStatement. Or are you suggesting a new Controller Class TIPPCoverageStatementController that delivers TIPPCoverageStatment-JSON to be inserted in a TIPP?

hornmo: Since the domain class TIPPCoverageStatement has its own `coverageDepth`, it also has a separate RefdataCategory `TIPPCoverageStatement.CoverageDepth` (which contains the same values as `TitleInstancePackagePlatform.CoverageDepth`). Since we return the Database-ID of the values as well and support updating the property by providing an ID, we should take care to output the correct values here. Therefore it should be 

`cat = RefdataCategory.findByLabel("TIPPCoverageStatement.CoverageDepth")`

VolkerHBZ: ok. done

hornmo (approved):

PR #308, opened 2020-08-04
#270 title extension

VolkerHBZ:
changed the *.gsps and added OAI Export of the new properties

### DIFF
--- a/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
+++ b/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
@@ -1393,6 +1393,17 @@ class IntegrationController {
    *    'editStatus':'the_publisher',
    *    'status':'the_publisher',
    *    'historyEvents':[
+   *    ],
+   *    'series':'the_publisher',
+   *    'subjectGroup':'the_publisher',
+   *    'prices':[
+   *      {
+   *       'priceType':'list',
+   *       'currency':'EUR',
+   *       'startDate':'yyyy-MM-dd',
+   *       'endDate':'yyyy-MM-dd',
+   *       'price':12.89
+   *      }
    *    ]
    *  }
    */
--- a/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
+++ b/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
@@ -1400,8 +1400,6 @@ class IntegrationController {
    *      {
    *       'type':'list',
    *       'currency':'EUR',
-   *       'startDate':'yyyy-MM-dd',
-   *       'endDate':'yyyy-MM-dd',
    *       'amount':12.89
    *      }
    *    ]
--- a/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
+++ b/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
@@ -1395,14 +1395,14 @@ class IntegrationController {
    *    'historyEvents':[
    *    ],
    *    'series':'the_publisher',
-   *    'subjectGroup':'the_publisher',
+   *    'subjectArea':'the_publisher',
    *    'prices':[
    *      {
-   *       'priceType':'list',
+   *       'type':'list',
    *       'currency':'EUR',
    *       'startDate':'yyyy-MM-dd',
    *       'endDate':'yyyy-MM-dd',
-   *       'price':12.89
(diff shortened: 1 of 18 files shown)


server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
```diff
@@ -1559,13 +1568,13 @@ class IntegrationController {
...
+                def pubFrom = GOKbTextUtils.completeDateString(titleObj.publishedFrom)
+                def pubTo = GOKbTextUtils.completeDateString(titleObj.publishedTo, false)
 
-                  title_changed |= ClassUtils.setDateIfPresent(pubFrom, title, 'publishedFrom')
-                  title_changed |= ClassUtils.setDateIfPresent(pubTo, title, 'publishedTo')
-                }
+                log.debug("Completed date publishedFrom ${titleObj.publishedFrom} -> ${pubFrom}")
+
+                title_changed |= ClassUtils.setDateIfPresent(pubFrom, title, 'publishedFrom')
+                title_changed |= ClassUtils.setDateIfPresent(pubTo, title, 'publishedTo')
```
hornmo, line 1577: Like the date props the new title String properties (`series` and `subjectArea`) should be updated around here as well. Instead of setDateIfPresent for that we would use

https://github.com/openlibraryenvironment/gokb/blob/0a2a234c16b61a5501bbcfda722c26ed67cef655/server/gokbg3/src/main/groovy/com/k_int/ClassUtils.groovy#L174

server/gokbg3/src/integration-test/groovy/gokbg3/rest/IdentifierTestSpec.groovy
```diff
@@ -44,6 +44,7 @@ class IdentifierTestSpec extends AbstractAuthSpec {
       accept('application/json')
     }
     then:
+//    resp.status == 200 // WRONG!
```
hornmo, line 47: Can be removed?

server/gokbg3/src/integration-test/resources/journal_prices_test.json
```diff
@@ -0,0 +1,67 @@
...
+      "subjectGroup": "Fringe",
+      "prices": [
+         {
+            "type": "list",
+            "currency": "EUR",
+            "amount": 123.45,
+            "startDate": "2010-01-31"
+         },
+         {
+            "typeype": "topup",
```
hornmo: Typo `typeype -> type`

server/gokbg3/src/integration-test/resources/journal_prices_test.json
```diff
@@ -0,0 +1,67 @@
...
+      "monographEdition": "null",
+      "editionDifferentiator": "null",
+      "publishedTo": "",
+      "editionNumber": "",
+      "publishedFrom": "1996-01-01 00:00:00.000",
+      "name": "Acta cytologica",
+      "continuingSeries": "",
+      "status": "Current",
+      "series": "Mystery Cloud",
+      "subjectGroup": "Fringe",
```
hornmo: `subjectGroup -> subjectArea` for consistency

server/gokbg3/src/integration-test/resources/journal_prices_test.json
```diff
@@ -0,0 +1,67 @@
...
+         {
+            "type": "eissn",
+            "value": "1938-2650"
+         },
+         {
+            "type": "ezb",
+            "value": "150346"
+         }
+      ],
+      "monographEdition": "null",
```
hornmo: Monograph fields may be omitted entirely for `type: "Serial"`. Otherwise `null` would probably be the desired value here instead of the String value.

hornmo (changes requested):

server/gokbg3/src/integration-test/groovy/gokbg3/rest/IdentifierTestSpec.groovy
VolkerHBZ: sometimes, the test returns a 200 status for no obvious reason. I suspect some html session to be alive still in the server for being responsible to validate the user.

server/gokbg3/src/integration-test/resources/journal_prices_test.json
VolkerHBZ: this is left from a previous stage of implementation, when that field was named subjectGroup...

Changed it.

hornmo (approved):

Issue #155, opened 2020-02-06
View Title History: Splitting/Merging Titles


dirk-ecker:
It is mentioned that titles could be splitted or merged, but not how this can be accomplished.

(p. 55)

darupp:
The splitting and merging is done via the buttons "Vorgänger" and "Nachfolger" that open the modal dialogue (p. 56). Regarding the data, this is defined by a title history event with 
  * set of predecessor title references
  * set of successor title references
  * event date

https://laser-qa.hbz-nrw.de/box/gokb-dbschema/tables/component_history_event.html

PR #310, opened 2020-08-10
Rest titles history

hornmo:
- Add endpoints for title histories (#155 )
- Add endpoint `/title-types`
- Enable title upsert via TIPPs in package POST


server/gokbg3/grails-app/controllers/org/gokb/rest/TitleController.groovy
```diff
@@ -27,8 +27,16 @@ class TitleController {
   def messageService
   def restMappingService
   def titleLookupService
+  def titleHistoryService
   def componentLookupService
 
+  @Secured(['ROLE_USER', 'IS_AUTHENTICATED_FULLY'])
+  def getTypes() {
+    def result = ["serial","monograph","database"]
+
```
VolkerHBZ, line 36: I'd rather prefer to extract the values from the database, as this is where they reside. If changes appear (new types of media), there will be no need to adapt the implementation and the change becomes effective immediately.

hornmo: You mean checking for existing subclasses of TitleInstance and returning their class name as well as their `niceName`?

VolkerHBZ: no, you'e right... I thought, it's a refdata value, like coverage-depth. Since new types of media will implement new subclasses anyway, this is a sufficient solution.

VolkerHBZ (approved):

PR #317, opened 2020-09-02
wrapping the response provider object in a data field

### LINKED ISSUE #296: View Provider: GET Response not nested in data
The response of getting a provider is not nested inside a data object analogue to the other services.

VolkerHBZ:


hornmo (commented):
For endpoints predictably returning a single object (i.e. a request with a supplied `id`), the client currently does not expect the response to contain a data container.

VolkerHBZ:
not needed

Issue #220, opened 2020-03-26
View Search Packages: GOKb-Format/KBART-Format not specified


dirk-ecker:
It is a formal specification needed to implement the export in GOKb-Format/KBART-Format.

darupp:
You can find the specification linked here: https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=470024321

- The first link leads to the official KBART documentation
- At the bottom of the table, you find proprietary extensions of the GOKb

dirk-ecker:
Thanks for the link.
The filename format should be 

[ProviderName]_[Region/Consortium]_[PackageName]_[YYYY-MM-DD].txt

but in the near future we will have the possibility to choose more than one provider. Consortium means the same as curatory group? Where to get the PackageName? Do we need a dialog to specify these parameters?

In other words, what should the format be?

darupp:
There will still be only one provider per package, see pp. 35/36 or pp. 49/50. The export files are always package specific. If in the list on p. 49 more than one package is picked, there should be multiple KBART and GOKb files exported (maybe in a ZIP). 

All String fields in the table should be normalized regarding the KBART title to fit the filename format (All blanks should be underscores, no forbidden characters etc.).

The [ProviderName] should be the name of the provider shown in the list on p.  ( @hornmo and I thought about using an short provider alternate name tagged as abbreviation. What do you think, @hornmo?)

The [PackageName] should be the name in the shown list.

The [Region/Consortium] is right now per convention part of the [PackageName], so there ist now further specification needed. 

I don't think a dialog is needed.

hornmo:
There are existing public endpoints for exporting Packages which can be used:

KBART: <gokb-url>/packages/kbart/<uuid|OID>
GOKb-Format: <gokb-url>/packages/packageTSVExport/<uuid|OID>

These Endpoints return *.tsv files. We are planning to improve the parameter handling to also work with the database id.

dirk-ecker:
Okay, good point. We should use these, instead of creating new one, I think.

dirk-ecker:
Nice.
I've tested the following url

https://gokbt.gbv.de/gokb/packages/kbart/6196e907-59df-4cfd-94f6-91dd1e72b0b0

It gives me a 403 and Invalid CORS request.

Can you adjust the service to accept my request?

We should extend the service so that it can handle a query to specify the packages to export.

@darupp At least we need a select to choose the format KBART or GOKb and also a progress indicator.

[...]

PR #326, opened 2020-09-29
#220 kbart export extension

VolkerHBZ:


hornmo (changes requested):
Functionality is fine, but the naming of KBART files need some work. I would recommend using the regular `name` of Package & Provider and manually simplifying them via replaceAll (/[:\s]/ or something similar) to get to the needed form (provider_name_date).

hornmo (approved):

PR #332, opened 2020-10-06
Rest tipps crossref

hornmo:
- Add PUT/PATCH REST endpoints for Sources
- Add POST endpoint for ReviewRequests
- Move `IntegrationController.processHistoryEvents` to TitleHistoryService
- Move `IntegrationController.addPublisherHistory` to TitleLookupService
- Catch `MultipleComponentsMatchedException` in IntegrationController actions
- Add UpdateToken generation for `PackageController.update()`
- Catch validation errors for title generation in `PackageController.save()`
- Extend TIPP coverageDepth validation/upsert
- Add PUT endpoint for package TIPPs (similar to `IntegrationController.crossReferencePackage`)
- Fix typo in ReviewsController
- Add allocatedGroups to ReviewRequest default embeds
- Add ReviewRequest allocatedGroups handling for ComponentLookupService
- Add type filter in GlobalController
- Fix CuratoryGroup name validation
- Add flexible handling for RefdataValues in `RestMappingServive.updateObject` & `ClassUtils.setRefdataIfPresent`

### DIFF
--- a/server/gokbg3/grails-app/controllers/gokbg3/UrlMappings.groovy
+++ b/server/gokbg3/grails-app/controllers/gokbg3/UrlMappings.groovy
@@ -17,6 +17,7 @@ class UrlMappings {
       post "/packages/$id/retire"(controller: 'package', namespace: 'rest', action:'retire')
       post "/packages/$id/tipps"(controller: 'package', namespace: 'rest', action: 'addTipps')
       put "/packages/$id/tipps"(controller: 'package', namespace: 'rest', action: 'updateTipps')
+      patch "/packages/$id/tipps"(controller: 'package', namespace: 'rest', action: 'updateTipps')
       "/packages/$id/$action"(controller: 'package', namespace: 'rest')
       get "/packages/$id"(controller: 'package', namespace: 'rest', action: 'show')
       put "/packages/$id"(controller: 'package', namespace: 'rest', action: 'update')
--- a/server/gokbg3/grails-app/controllers/gokbg3/UrlMappings.groovy
+++ b/server/gokbg3/grails-app/controllers/gokbg3/UrlMappings.groovy
@@ -97,6 +97,8 @@ class UrlMappings {
 
       get "/sources"(controller: 'sources', namespace: 'rest', action: 'index')
       get "/sources/$id"(controller: 'sources', namespace: 'rest', action: 'show')
+      put "/sources/$id"(controller: 'sources', namespace: 'rest', action: 'update')
+      patch "/sources/$id"(controller: 'sources', namespace: 'rest', action: 'update')
       post "/sources"(controller: 'sources', namespace: 'rest', action: 'save')
 
       get "/reviews"(controller: 'reviews', namespace: 'rest', action: 'index')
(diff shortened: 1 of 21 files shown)


server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
```diff
@@ -203,6 +203,13 @@ class BootStrap {
       }
       log.debug("${ctr} components updated");
 
+    log.info("GoKB remove usused refdata");
+      def rr_std = RefdataCategory.lookup('ReviewRequest.StdDesc', 'RR Standard Desc 1')
+
+      if (rr_std) {
+        rr_std.delete()
```
VolkerHBZ, line 210: Cleaning known junk at startup seems weird. Where does this RefdataCategory come from in the first place? Why it's written to the database at all and why isn't that avoided?

server/gokbg3/src/main/groovy/com/k_int/ClassUtils.groovy
```diff
@@ -152,7 +152,7 @@ class ClassUtils {
         v = RefdataCategory.lookup(cat, value)
       }
     }
-    else if (value && cat) {
+    else if (value instanceof Integer && cat) {
       try {
         def candidate = RefdataValue.get(value)
```
VolkerHBZ, line 157: Not sure, but afaik get() works on Database ids of type Long, not Integer.

server/gokbg3/grails-app/controllers/gokbg3/UrlMappings.groovy
```diff
@@ -16,6 +16,7 @@ class UrlMappings {
     group "/rest", {
       post "/packages/$id/retire"(controller: 'package', namespace: 'rest', action:'retire')
       post "/packages/$id/tipps"(controller: 'package', namespace: 'rest', action: 'addTipps')
+      put "/packages/$id/tipps"(controller: 'package', namespace: 'rest', action: 'updateTipps')
       "/packages/$id/$action"(controller: 'package', namespace: 'rest')
```
VolkerHBZ, line 21: just add `patch "/packages/$id/tipps"(controller: 'package', namespace: 'rest', action: 'updateTipps')` as well

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
hornmo: The value was added via BootStrap.groovy before, which was only removed in a previous patch. We can therefore assume it was at some point created in all existing instances so has to be removed everywhere.

server/gokbg3/src/main/groovy/com/k_int/ClassUtils.groovy
hornmo: It does work with both actually. GORM defines the type as `java.io.Serializable` internally.

VolkerHBZ (approved):

server/gokbg3/grails-app/controllers/gokbg3/UrlMappings.groovy
hornmo: Agreed. PATCH now has the behaviour of `addOnly` by default, where missing TIPPs are simply ignored.

PR #334, opened 2020-10-12
#327 source in package

VolkerHBZ:


server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -375,6 +377,7 @@ class FTUpdateService {
         result.tippPackage = kbc.pkg ? kbc.pkg.getLogEntityId() : ""
         result.tippPackageName = kbc.pkg ? kbc.pkg.name : ""
         result.tippPackageUuid = kbc.pkg ? kbc.pkg.uuid : ""
+        result.tippPackageSource = kbc.pkg.source ? kbc.pkg.source : ""
```
hornmo: Not sure if we need the package source on each single TIPP as well. Is there a specific use-case?

server/gokbg3/grails-app/domain/org/gokb/cred/Source.groovy
```diff
@@ -29,8 +28,7 @@ class Source extends KBComponent {
     defaultAccessURL(nullable:true, blank:true)
     explanationAtSource(nullable:true, blank:true)
     contextualNotes(nullable:true, blank:true)
-    frequency(nullable:true, blank:true)
-
+    frequency(nullable:true, blank:true, default: "d")
```
hornmo: According to the UI spec this would have to be `1D`.

server/gokbg3/grails-app/domain/org/gokb/cred/Package.groovy
```diff
@@ -480,6 +480,13 @@ select tipp.id,
             'name'(provider.name)
           }
         }
+
+        if (source) {
+          builder.'source'([id: source.id, uuid: source.uuid]) {
+            'name'(source.name)
+          }
+        }
+
```
hornmo: Since Source information is a general KBComponent link, this information is already added at the start via `addCoreGOKbXmlFields`.

scripts/es_reset.sh
```diff
@@ -180,6 +180,13 @@ curl -X PUT "localhost:9200/$INDEXNAME/component/_mapping" -H 'Content-Type: app
           "value": { "type": "keyword"}
         }
       },
+      "source" : {
+        "type" : "nested",
+        "properties": {
+          "name": { "type": "keyword"},
+          "url": { "type": "keyword"}
+        }
+      },
```
hornmo, line 189: Since we will generate the `name` of the Source directly from the Package name, we won't really need this field here. More useful would probably be to add the `frequency` instead.

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -85,6 +85,9 @@ class FTUpdateService {
         result.scope = kbc.scope ? kbc.scope.value : ""
         result.listVerifiedDate = kbc.listVerifiedDate ? sdf.format(kbc.listVerifiedDate) : ""
 
+        if (kbc.source)
+          result.source = [name : kbc.source.name, id:kbc.source.id]
```
hornmo: Adjust according to requested index mapping changes (`frequency` instead of `name`)

hornmo (changes requested):

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
VolkerHBZ: no specific use case, just storing the source along with the package for later use by external entities, as this was demanded by the sentence:"Die Daten müssen entsprechend beim Package auch über die APIs ausgeliefert werden (OAI-PMH und ES)". Implementing it this way was my interpretation of that line.

I'll erase it. We can put it back, if needed.

server/gokbg3/grails-app/domain/org/gokb/cred/Source.groovy
VolkerHBZ: ok.

server/gokbg3/grails-app/domain/org/gokb/cred/Package.groovy
VolkerHBZ: true. i didn't notice that before. thx.

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
VolkerHBZ: done

hornmo (approved):

Issue #300, opened 2020-08-03
View Provider: Improve service to search for identifier


dirk-ecker:
Suggesting an identifier for searching a provider should be improved, so that only identifiers of providers are shown.
Therefore we need a parameter..

VolkerHBZ:
Since there's no direct link from providers to identifiers, I suppose, you expect something like "return all identifiers of all publications by the given provider"!?

dirk-ecker:
Yes that would it be I think. The reason to request this enhancement is to only get available ids, but I do not know the user intention. Does it makes sense?

VolkerHBZ:
the endpoint /rest/identifier-namespaces now has a parameter to specify the type of namespaces to list:
`/rest/identifier-namespaces?targetType=<Org|Package|Title|Book|Journal|Database|Other>`
There is no special TargetType for Providers or Publishers, as they're all Orgs. If it's impossible to use only Org as parameter value in the request, the implementation can be broadened to map any other string to Org.

dirk-ecker:
Okay, this tickets needs further improvements. During implementation I saw that you are talking about namespace and I instead mean identifier. I want to search for identifiers for a specific target type not for  namespaces. So it should be `/rest/identifier?...`.

By the way I found that the latter gives also the namespaces.

PR #337, opened 2020-10-15
#300 typed namespaces

VolkerHBZ:
titleNamespace and packageNamespace as properties in Org


server/gokbg3/grails-app/domain/org/gokb/cred/Org.groovy
```diff
@@ -203,17 +206,19 @@ class Org extends KBComponent {
...
-
-        if ( roles ) {
+
+    builder.'gokb'(attr) {
+      builder.'org'(['id': (id), 'uuid': (uuid)]) {
+
+        addCoreGOKbXmlFields(builder, attr)
+        builder.'homepage'(homepage)
+        builder.'titleNamespace'('namespaceName': titleNamespace.name, 'value': titleNamespace.value, 'id': titleNamespace.id)
+        builder.'packageNamespace'('namespaceName': packageNamespace.name, 'value': packageNamespace.value, 'id': packageNamespace.id)
```
hornmo: Needs checks for null state. Missing namespaces should result in an empty element.

hornmo (changes requested):

server/gokbg3/grails-app/domain/org/gokb/cred/Org.groovy
VolkerHBZ: done

hornmo (approved):

PR #353, opened 2020-11-02
346 package global note

darupp:
Added test field `globalNote` to `Package` to specify the range in case of a global Setting of "Consortium", "Region" or "Other"


server/gokbg3/src/integration-test/groovy/gokbg3/rest/PackageTestSpec.groovy
```diff
@@ -171,6 +171,7 @@ class PackageTestSpec extends AbstractAuthSpec {
       descriptionURL : "https://heise.de",
       fixed          : "Yes",
       global         : "Consortium",
+      globalNote     : "Testing Consortium",
```
VolkerHBZ, line 174: there's no testing for the new property here. 
add something like     resp.json.globaleNote == "Testing Consortium" in the then: section of the test method

darupp: Added the test results

hornmo (approved):

PR #359, opened 2020-11-10
automatic triggering of Ygor Package Updates 

### LINKED ISSUE #360: Manueller Update-Service über URL in "Available Actions"
- In der alten GUI unter Packageview unter "Available Action" die Funktion "Manuelles Update jetzt" 
- Keine Vorab-Prüfung, ob Source, URL, Service etc. existieren, sondern Fehlermeldung danach (<- Einfacher zu Implementieren)
- Evtl. noch Probleme dadurch, dass mehrere Paketjobs zurzeit parallel laufen können.

### LINKED ISSUE #361: Anlegen der Source über die Package Crossref
Über die Package Crossref soll eine Source angelegt werden können
- Wenn die ID fehlt, dann soll sie angelegt werden
- Namen nicht mitgeschickt, da er ja in der GOKb gleich des Paketnamens gesetzt werden soll
- Felder: In JSON wie Domänenattribute benennen
  - `url`
  - `freqency` (Siehe Pull-Request für Syntax) --> 1d ist GOKb-Seitig standard, muss nicht mitgeschickt werden.
  - `automaticUpdates` BOOLEAN
  - `zdbMatch` BOOLEAN
  - `ezbMatch` BOOLEAN
  - `targetNamespace`: ID des nachgeschlagenen Namespace

VolkerHBZ:
at 6:00 am when
config.gokb.packageUpdate_enabled = true and
config.gokb.ygorUrl is set

### DIFF
--- a/server/gokbg3/grails-app/domain/org/gokb/cred/Source.groovy
+++ b/server/gokbg3/grails-app/domain/org/gokb/cred/Source.groovy
@@ -1,6 +1,8 @@
 package org.gokb.cred
 
 import javax.persistence.Transient
+import java.util.regex.Matcher
+import java.util.regex.Pattern
 
 class Source extends KBComponent {
 
@@ -141,5 +143,38 @@ class Source extends KBComponent {
         }
       }
     }
+  }
+
+  public boolean needsUpdate() {
+    Source src = this
+    if (src.lastRun == null) {
+      return true;
+    }
+    if (src.frequency != null) {
+      Pattern pattern = Pattern.compile("\\s*(\\d*)\\s*([a-zA-Z]*)\\s*");
+      Matcher matcher = pattern.matcher(src.frequency);
+      int days
+      if (matcher.find()) {
+        def length = [
+          D: 1,
+          T: 1,
+          W: 7,
+          M: 30,
+          Q: 91,
+          H: 182,
+          J: 365,
+          Y: 365]
+        def interval = matcher.group(2)
+        def number = matcher.group(1)
+        days = Integer.parseInt(number) * length[interval.toUpperCase()]
+
+        Date today = new Date()
+        Date due = src.lastRun.plus(days)
+        if (due.before(today)) {
+          return true
+        }
+      }
+    }
+    return false
   }
 }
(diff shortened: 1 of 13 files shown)


hornmo:
The jobs should also consider the defined frequency of each source. The new config items should be added with default values (false & null) to the global config in application.yml.

server/gokbg3/grails-app/domain/org/gokb/cred/Source.groovy
```diff
@@ -142,4 +144,37 @@ class Source extends KBComponent {
...
+    if (src.lastRun == null) {
+      return true;
+    }
+    if (src.frequency != null) {
+      Pattern pattern = Pattern.compile("\\s*(\\d*)\\s*([a-zA-Z]*)\\s*");
+      Matcher matcher = pattern.matcher(src.frequency);
+      int days
+      if (matcher.find()) {
+        def length = [
+          D: 1,
```
philboeselager, line 159: This is OK for now, but would need to be refined. Chosen the option "M1" for example, the user would expect the update to happen at the very same day of each month.

server/gokbg3/src/main/groovy/com/k_int/ClassUtils.groovy
```diff
@@ -197,4 +197,14 @@ class ClassUtils {
       }
     return false
   }
+
+  public static boolean setBooleanIfDifferent(obj, prop, value) {
+    if ((obj != null) && (prop != null) && (value != null)) {
+      if (obj[prop] != value) {
```
philboeselager, line 203: The check in 203 is not necessary.

philboeselager (approved):
The intervals table (first comment) should be refined later.
The unnecessary check (second comment) can be removed at next occasion.

server/gokbg3/grails-app/domain/org/gokb/cred/Source.groovy
VolkerHBZ: M1 won't pass the parser, as it expects the number *before* the letter. Maybe the user expects the update at the same day-of-month, though I highly doubt that he or she notices the slight shift of 2 days max, within one month. There's no way of entering that specific day-of-month for the update anyway. The best solution might actually be to use your _inverted_ syntax, with "M1" meaning each first day of the month or "W3" for the thirds day each week.

PR #385, opened 2021-01-05
380 elasticsearch scroll

philboeselager:


server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
```diff
@@ -423,99 +422,86 @@ class ESSearchService{
...
+      SearchScrollRequest scrollRequest = new SearchScrollRequest(params.scrollId)
+      scrollRequest.scroll("1m")
+      response = esClient.searchScroll(scrollRequest)
+    }
+    log.debug("scrollId : " + response.actionGet().getScrollId())
+    result.scrollId = response.actionGet().getScrollId()
+    SearchHit[] searchHits = response.actionGet().getHits().getHits()
+    result.records = []
+    for (SearchHit hit in searchHits){
+      result.records << hit.getSourceAsString()
```
hornmo: Why do we want the records as strings instead of objects?

philboeselager: @hornmo True. Now being returned as a map.

hornmo (approved):

PR #393, opened 2021-01-12
Es update

hornmo:
- Bump internally used ES version to 6.8.13
- Switch from deprecated TransportClient to REST

This is  compatible with indices created under 5.X, so this won't necessitate a full reindexing. Upgrading this to 7.X contains breaking changes which will require changing the index mapping and several retrieval & update methods.


server/gokbg3/grails-app/controllers/org/gokb/GlobalSearchController.groovy
```diff
@@ -43,13 +45,12 @@ class GlobalSearchController {
...
         log.debug("Using index ${grailsApplication.config.gokb?.es?.index /*?: 'gokbg3dev (auto)'*/}")
 
-        SearchRequestBuilder es_request = esclient.prepareSearch("globalSearch")
-            .setIndices(grailsApplication.config.gokb?.es?.index ?: "gokbg3")
-            .setTypes(grailsApplication.config.globalSearch.types ?: "component")
-            .setSize(result.max)
-            .setFrom(result.offset)
-            .setQuery(esQuery)
-            .addAggregation(
+        SearchRequest es_request = new SearchRequest(grailsApplication.config.gokb?.es?.index ?: "gokbg3")
```
philboeselager, line 48: I'm not really a fan of hard-coded fallback names. An installation might use another index name, where this leads to exceptions. I suggest to create another ticket for this. Shall I?

philboeselager (approved):
Still needs adjustment of `changedSince`, which will be included on https://github.com/openlibraryenvironment/gokb/tree/380-elasticsearchScroll .

philboeselager (approved):

PR #400, opened 2021-01-21
378 add fields to elasticsearch

philboeselager:
@hornmo Still not sure if we should add mappings to `es_reset.sh` - please let me know if you think this must or should be done. In my tests, this worked without any problem.

### DIFF
--- a/server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
+++ b/server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
@@ -443,23 +443,21 @@ class FTUpdateService {
         }
 
         // title history for all title types
-        if (kbc.title?.hasProperty("historyEvents")){
-          result.titleHistory = []
-          kbc.title?.titleHistory?.each{ he ->
-            if (he.date){
-              def event = [:]
-              event.date = dateFormatService.formatIsoTimestamp(he.date)
-              event.from = []
-              if (he.from) {
-                event.from.addAll(he.from.collect { fe -> [id: fe.id, uuid: fe.uuid, name: fe.name ]})
-              }
-              event.to = []
-              if (he.to){
-                event.to.addAll(he.to.collect{ te -> [id: te.id, uuid: te.uuid, name: te.name] })
-              }
-              event.id = he.id ?: ""
-              result.titleHistory.add(event)
+        result.titleHistory = []
+        kbc.title?.titleHistory?.each{ he ->
+          if (he.date){
+            def event = [:]
+            event.date = dateFormatService.formatIsoTimestamp(he.date)
+            event.from = []
+            if (he.from) {
+              event.from.addAll(he.from.collect { fe -> [id: fe.id, uuid: fe.uuid, name: fe.name ]})
             }
+            event.to = []
+            if (he.to){
+              event.to.addAll(he.to.collect{ te -> [id: te.id, uuid: te.uuid, name: te.name] })
+            }
+            event.id = he.id ?: ""
+            result.titleHistory.add(event)
           }
         }
 
--- a/server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
+++ b/server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
@@ -389,6 +389,74 @@ class FTUpdateService {
         }
 
         result.componentType = kbc.class.simpleName
+        result.tippTitleMedium = kbc.title ? kbc.title.medium : ""
+
+        result.accessStartDate = kbc.accessStartDate ? kbc.accessStartDate.toString() : ""
+        result.accessEndDate = kbc.accessEndDate ? kbc.accessEndDate.toString() : ""
+
+        result.subjectArea = kbc.subjectArea ?: ""
+        result.series = kbc.series ?: ""
+
+        if (kbc.title?.niceName == 'Book'){
+
+          // edition for eBooks
+          def edition = [:]
+          if (kbc.title?.editionNumber){
+            edition.number = kbc.title.editionNumber
+          }
+          if (kbc.title?.editionDifferentiator){
+            edition.differentiator = kbc.title.editionDifferentiator
+          }
+          if (kbc.title?.editionStatement){
+            edition.statement = kbc.title.editionStatement
+          }
+          if (!edition.isEmpty()){
+            result.titleEdition = edition
+          }
+
+          // simple eBook fields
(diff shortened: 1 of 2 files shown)


server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -389,6 +398,76 @@ class FTUpdateService {
...
+          publisher.uuid = pub.uuid ?: ""
+          result.titlePublishers.add(publisher)
+        }
+
+        // prices for all title types
+        result.prices = []
+        kbc.prices?.each { p ->
+          def price = [:]
+          price.type = p.priceType?.value ?: ""
+          price.amount = p.price ?: ""
```
hornmo: Since `price` is a Float, the field type would be ambiguous if created dynamically. Would converting it to String be fine?

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -389,6 +398,76 @@ class FTUpdateService {
...
+          price.currency = p.currency ?: ""
```
hornmo: Should be `currency.value` (RefdataValue property).

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -389,6 +398,76 @@ class FTUpdateService {
...
+          price.startDate = p.startDate ?: ""
+          price.endDate = p.endDate ?: ""
```
hornmo: Same as access dates, please use `dateFormatService.formatIsoTimestamp(Date date)`.

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -389,6 +398,76 @@ class FTUpdateService {
         }
 
         result.componentType = kbc.class.simpleName
+        result.tippTitleMedium = kbc.title ? kbc.title.medium : ""
+
+        result.accessStartDate = kbc.accessStartDate ? kbc.accessStartDate.toString() : ""
+        result.accessEndDate = kbc.accessEndDate ? kbc.accessEndDate.toString() : ""
```
hornmo: We are using a centralized DateFormatService to format Date objects. In this case `dateFormatService.formatIsoTimestamp(Date date)` would be the way to go.

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -389,6 +398,76 @@ class FTUpdateService {
...
+            edition.statement = kbc.title.editionStatement
+          }
+          if (!edition.isEmpty()){
+            result.titleEdition = edition
+          }
+
+          // simple eBook fields
+          result.titleVolumeNumber = kbc.title?.volumeNumber ?: ""
+          result.titleDateFirstInPrint = kbc.title?.dateFirstInPrint ?: ""
+          result.titleDateFirstOnline = kbc.title?.dateFirstOnline ?: ""
```
hornmo: Same as access dates, please use dateFormatService.formatIsoTimestamp(Date date).

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -389,6 +398,76 @@ class FTUpdateService {
...
+          result.titleFirstEditor = kbc.title?.firstEditor ?: ""
+          result.titleFirstAuthor = kbc.title?.firstAuthor ?: ""
+          result.titleImprint = kbc.title?.imprint ?: ""
```
hornmo: An imprint of a title would be an Imprint class object, so `ìmprint?.name` would be better.

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -389,6 +398,76 @@ class FTUpdateService {
...
+        }
+
+        // title history for all title types
+        if (kbc.title?.hasProperty("historyEvents")){
+          result.titleHistory = []
+          kbc.title?.historyEvents?.each{ he ->
```
hornmo: `TitleInstance.historyEvents` is not defined. `title.titleHistory` returns a collection of directly connected history events for the title. `from` and `to` are arrays of TitleInstances, which would have to be mapped here as well. Events have an `id` but no `uuid`. The event `event.date` should be converted via the service as well.

hornmo (changes requested):

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -438,10 +440,12 @@ class FTUpdateService {
...
-              event.uuid = he.uuid ?: ""
+              event.date = dateFormatService.formatIsoTimestamp(he.date)
+              event.from = []
+              event.from.addAll(he.from)
+              event.to = []
+              event.to.addAll(he.to)
+              event.id = he.id ?: ""
               result.titleHistory.add(event)
             }
           }
```
hornmo: `title.historyEvents` should be replaced with `title.titleHistory`. Simply doing `addAll` is not not enough here, since `from` & `to` are filled with TitleInstance objects, which cannot be automatically serialized in JSON. Therefore, every entry in `from`/`to` has be to manually converted. Something like `he.from.collect { fe -> [id: fe.id, uuid: fe.uuid, name: fe.name ]}` would probably be the shortest solution.

philboeselager: Thanks for explaining again! Commit is going to follow.

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -437,14 +437,18 @@ class FTUpdateService {
         // title history for all title types
         if (kbc.title?.hasProperty("historyEvents")){
```
hornmo: `historyEvents` is not defined. Just removing the check would be fine.

hornmo (changes requested):

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
philboeselager: True. This is a relict, and non-functional as well.

hornmo (approved):

PR #402, opened 2021-01-27
24 add es package field

philboeselager:


hornmo (approved):
Missing later changes from other commit. Another merge

PR #405, opened 2021-01-29
X ref service 368

VolkerHBZ:
Refactored crossReferencePackage process, encapsulating one run in a separate class. Might be a template for other processes.


philboeselager (commented):
Code-wise, this looks fine. Functionally, I cannot evaluate this.
We should reduce the size of PR's by commiting to a defined code style.
Re-formatting the code to this defined code style should be done commit-apart from any development.

hornmo (approved):

PR #410, opened 2021-02-08
Tipp validierung #14

VolkerHBZ:
moved validateDTO in the respective Domain classes
added some new media types


server/gokbg3/grails-app/domain/org/gokb/cred/TitleInstance.groovy
```diff
@@ -627,122 +627,194 @@ class TitleInstance extends KBComponent {
    *   type:'Serial' or 'Monograph'
    *}*/
   @Transient
-  public static def validateDTO(titleDTO) {
-    def result = ['valid': true, 'errors': [:]]
+  public static def validateDTO(JSONObject titleDTO) {
+    def result = ['valid': true]
+    def valErrors = []
```
hornmo: This does not conform to the current object error layout. `result.errors` of any validateDTO should be a Map with property names as keys (same as `messageService.processValidationErrors()`).

server/gokbg3/grails-app/domain/org/gokb/cred/TitleInstance.groovy
```diff
@@ -627,122 +627,194 @@ class TitleInstance extends KBComponent {
...
+          return RefdataCategory.lookupOrCreate("TitleInstance.Medium", "Other")
+        default:
+          return null
+      }
+    }
+    else {
+      return null
+    }
+  }
+
```
hornmo: Since the values are all created in Bootstrap, `RefdataCategory.lookup(<cat>, <value>)` would be better here.

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
```diff
@@ -534,14 +537,10 @@ class BootStrap {
...
-    RefdataCategory.lookupOrCreate("TitleInstance.Medium", "Audio").save(flush: true, failOnError: true)
-    RefdataCategory.lookupOrCreate("TitleInstance.Medium", "Book").save(flush: true, failOnError: true)
-    RefdataCategory.lookupOrCreate("TitleInstance.Medium", "Database").save(flush: true, failOnError: true)
-    RefdataCategory.lookupOrCreate("TitleInstance.Medium", "Dataset").save(flush: true, failOnError: true)
-    RefdataCategory.lookupOrCreate("TitleInstance.Medium", "Film").save(flush: true, failOnError: true)
-    RefdataCategory.lookupOrCreate("TitleInstance.Medium", "Image").save(flush: true, failOnError: true)
-    RefdataCategory.lookupOrCreate("TitleInstance.Medium", "Journal").save(flush: true, failOnError: true)
+    ["A & I Database", "Audio", "Book", "Database", "Dataset", "Film", "Image", "Journal",
+     "Other", "Published Score", "Article", "Software", "Statistics", "Market Data", "Standards",
+     "Biography", "Legal Text", "Cartography", "Miscellaneous", "Other"].each { med ->
```
hornmo: List contains `Other` twice.

hornmo (changes requested):
A consistent way to unify this properly would be to refer everything back to the index of the specific TIPP to basically bundle errors for the specific KBART line, and put everything in the same hierarchy as the input format. So `result.errors.tipps` would be an array like: 

```
[
    [
        index: 1456,
        errors: [
            title: [
                name: [
                    [message: "No title name", baddata: <tipp.title.name>],
                ],
                publishedFrom: [
                   [message: "Illegal date for publishedFrom", baddata: <tipp.title.publishedFrom>]
                ]
           ],
          url: [
               [message: "Illegal URL format", baddata: <tipp.url>]
           ]
      ]
  ],
]
```
So errors.tipps.size() would return the number of faulty KBART lines, and inside their `errors` object the structure would be the same as the one that came in.

server/gokbg3/grails-app/domain/org/gokb/cred/TitleInstance.groovy
VolkerHBZ: nope, as there are also german keywords that need to be mapped to the same refdataValues.

server/gokbg3/grails-app/domain/org/gokb/cred/TitleInstance.groovy
VolkerHBZ: the error objects are adapted according to the suggested layout.

hornmo (approved):

PR #413, opened 2021-02-10
32 es api tunnel

philboeselager:
Add a new Api endpoint serving Elasticsearch query language. Only open for GET requests. Thereby, the full query language of ES is available, withour needing to open the ES port, which can be dangerous in terms of index manipulation.


philboeselager:
Please merge the commits of Volker first, or cherry-pick the last commit.

PR #414, opened 2021-02-10
30 selectable frequency

philboeselager:
Make the Source frequency selectable by UI (Daily, Weekly, Monthly, Quarterly, Yearly).

### DIFF
--- a/server/gokbg3/grails-app/views/apptemplates/_source.gsp
+++ b/server/gokbg3/grails-app/views/apptemplates/_source.gsp
@@ -12,12 +12,14 @@
 				&nbsp;<a href="${d.url}" target="new">Follow Link</a>
 			</g:if>
 		</dd>
+
 		<dt>
 			<g:annotatedLabel owner="${d}" property="frequency">Frequency</g:annotatedLabel>
 		</dt>
 		<dd>
-			<g:xEditable class="ipe" owner="${d}" field="frequency" />
+			<g:xEditableRefData owner="${d}" field="frequency" config='Source.Frequency' />
 		</dd>
+
 		<dt>
 			<g:annotatedLabel owner="${d}" property="defaultSupplyMethod">Default Supply Method</g:annotatedLabel>
 		</dt>
--- a/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
+++ b/server/gokbg3/grails-app/controllers/org/gokb/IntegrationController.groovy
@@ -1264,9 +1264,7 @@ class IntegrationController {
             componentUpdateService.ensureCoreData(title, titleObj, fullsync, user)
 
             title_changed |= componentUpdateService.setAllRefdata([
-                'OAStatus', 'medium',
-                'pureOA', 'continuingSeries',
-                'reasonRetired'
+                'OAStatus', 'medium', 'pureOA', 'continuingSeries', 'reasonRetired'
             ], titleObj, title)
 
             def pubFrom = GOKbTextUtils.completeDateString(titleObj.publishedFrom)
--- a/server/gokbg3/grails-app/domain/org/gokb/cred/Source.groovy
+++ b/server/gokbg3/grails-app/domain/org/gokb/cred/Source.groovy
@@ -12,7 +12,7 @@ class Source extends KBComponent {
   String contextualNotes
   // Org combo -- What organisation - aggregator -- responsibleParty
   Boolean automaticUpdates = false
-  String frequency
+  RefdataValue frequency
   String ruleset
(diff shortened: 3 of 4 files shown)


server/gokbg3/grails-app/views/apptemplates/_source.gsp
```diff
@@ -12,12 +12,14 @@
 				&nbsp;<a href="${d.url}" target="new">Follow Link</a>
 			</g:if>
 		</dd>
+
 		<dt>
 			<g:annotatedLabel owner="${d}" property="frequency">Frequency</g:annotatedLabel>
 		</dt>
 		<dd>
-			<g:xEditable class="ipe" owner="${d}" field="frequency" />
+			<g:xEditableRefData owner="${d}" field="frequency" config='Source.Frequency' />
```
hornmo, line 20: `xEditableRefdata` is incompatible with setting a String property (like `frequency`), it is used to link two objects. Setting a String field via the `RefdataValue.value` would have to be handled by a new separate tag.

Generally properties with a controlled value range should just be of type RefdataValue, alternatively a custom validator for the specific String field would be in order. The issue of handling (now invalid) values of existing objects is not covered here though.

hornmo (changes requested):

server/gokbg3/grails-app/views/apptemplates/_source.gsp
philboeselager: @hornmo That's interesting. Actually, I did implement a new tag that was quite finished - only thing it was missing was the saving the field in the modelled object. What's the technique used for saving in the other tags? I haven't figured it out, yet.

At this moment, @VolkerHBZ came up with the idea to use `xEditableRefData`. I implemented it, and it works, right? Then why is it not appropriate to use it here?

VolkerHBZ: @hornmo changing the Type of the Source.frequency property from String to RefdataValue is the objective of this change. Therefore, I suggested to use the xEditableRefData tag for editing it - and: it works.

darupp: I think the misunderstanding is that `frequency`  has been refactored from `String` to a `RefdataValue` in [this issue](https://jira.hbz-nrw.de/browse/GOKBIT-30). Maybe these branches are not in sync?

philboeselager:
Changed type to RefdataValue, as this seems to be consent.

hornmo:
> 
> Changed type to RefdataValue, as this seems to be consent.

The type change in itself is alright. This will of course make the application _forget_ the String values of existing sources, just to be clear. I'm assuming this has been considered acceptable @darupp .

hornmo (approved):

philboeselager:
> > Changed type to RefdataValue, as this seems to be consent.
> 
> The type change in itself is alright. This will of course make the application _forget_ the String values of existing sources, just to be clear. I'm assuming this has been considered acceptable @darupp .

Forgetting the setting is not the intended behaviour of a settings UI element, right? So how can RefdataValue keep the setting, @hornmo , @VolkerHBZ ?

darupp:
> Forgetting the setting is not the intended behaviour of a settings UI element, right? So how can RefdataValue keep the setting, @hornmo , @VolkerHBZ ?

I think @hornmo meant that existing sources with given String values are not transferred to the new RefdataValue. This could be done via database script, but I think regarding the few sources that are given at the moment, it can be changed manually via the GUI.

Further, the new GUI has to be adapted to the datatype change.

hornmo:
> 
> 
> > > Changed type to RefdataValue, as this seems to be consent.
> > 
> > 
> > The type change in itself is alright. This will of course make the application _forget_ the String values of existing sources, just to be clear. I'm assuming this has been considered acceptable @darupp .
> 
> Forgetting the setting is not the intended behaviour of a settings UI element, right? So how can RefdataValue keep the setting, @hornmo , @VolkerHBZ ?

What is happening right now is column `frequency_id` (of type _bigint_) is created next to the existing varchar column `frequency` on deploy, and the application just starts to refer to the new column for the Source property. For preserving the existing values while using the same property name in the application so the application will also access them, we would need to do a db-migration which has to run before deployment of the new : 

- Create the new column manually
- Check for frequency values of existing Source objects
- Convert these values to their respective RefdataValue (which is not really possible for values like `5M`, so we will have to cut corners here)
- Write those new values in the new column
- Drop the old column when finished

An alternative would be to just have a new property (like _updateCycle_), which holds the new RefdataValue. We can then do the migration in Bootstrap.groovy, after which we might want to deprecate the usage of `frequency` going forward.

PR #415, opened 2021-02-15
Add new endpoint /api/elasticsearchTunnel/ (GOKbIT-32)

philboeselager:


server/gokbg3/grails-app/conf/application.groovy
```diff
@@ -82,6 +82,7 @@ grails.plugin.springsecurity.controllerAnnotations.staticRules = [
   [pattern: '/api/show',                access: ['ROLE_USER']],
   [pattern: '/api/namespaces',          access: ['permitAll']],
   [pattern: '/api/groups',              access: ['permitAll']],
+  [pattern: '/api/getApiTunnel',        access: ['permitAll']],
```
hornmo: The pattern has to refer to the controller action name rather than the service method

hornmo (changes requested):

server/gokbg3/grails-app/conf/application.groovy
philboeselager: That's right. Change follows.

hornmo (approved):

PR #419, opened 2021-02-18
adding ReviewRequests for wrong tipps

### LINKED ISSUE #383: Feedback zu nicht eingespielten Titeln / TIPPs
Im Zusammenhang eines Ygor-Tickets (siehe https://github.com/hbz/laser-ygor/issues/280#issuecomment-688179358) entstand der Verbesserungs-Vorschlag, bei der Einspielung abgelehnte TIPPs in einem Kommentar-Feld des Pakets aufzulisten, um dem Nutzer eine bessere Transparenz zu bieten.

Da es in diesem Zusammenhang offensichtlich Ygor-seitig nichts mehr zu tun gibt, und das Ticket etwas in Vergessenheit geraten zu scheint, überführe ich die Idee in dieses neue GOKb-Ticket.

VolkerHBZ:


server/gokbg3/grails-app/domain/org/gokb/cred/TitleInstance.groovy
```diff
@@ -767,24 +766,18 @@ class TitleInstance extends KBComponent {
     def type = null
 
     if (titleDTO.type) {
-      switch (titleDTO.type) {
+      switch (titleDTO.type.tolowerCase()) {
```
hornmo: `toLowerCase`

server/gokbg3/src/main/groovy/org/gokb/CrossRefPkgRun.groovy
```diff
@@ -150,7 +152,7 @@ class CrossRefPkgRun {
         }
         else {
           log.error("No package")
-          currentTippError.put(package:['message': messageService.resolveCode('crossRef.package.tipps.error.pkgId', [json_tipp.title.name], request_locale), baddata: json_tipp.package])
+          currentTippError.put(package: ['message': messageService.resolveCode('crossRef.package.tipps.error.pkgId', [json_tipp.title.name], request_locale), baddata: json_tipp.package])
```
hornmo: `package` should be the first parameter for put (`:` -> `,`)

server/gokbg3/src/main/groovy/org/gokb/CrossRefPkgRun.groovy
```diff
@@ -178,9 +180,23 @@ class CrossRefPkgRun {
           }
         }
 
-        if (currentTippError.size() > 1) {
+
+        if (invalidTipps.contains(json_tipp)) {
+          reviewRequestService.raise(
+            pkg,
+            "TIPP rejected",
+            "TIPP ${json_tipp.name ?: json_tipp.title.name} coudn't be imported. ${(currentTippError as JSON).toString()}",
```
hornmo, line 187: The error object in itself should probably go into the `additionalInfo` JSON field for later output processing, rather than adding it to a user-facing message String.

server/gokbg3/src/main/groovy/org/gokb/CrossRefPkgRun.groovy
```diff
@@ -434,8 +449,8 @@ class CrossRefPkgRun {
           def he_result = titleHistoryService.processHistoryEvents(ti, titleObj, title_class_name, user, fullsync, locale)
           if (he_result.errors) {
             currentTippError.put(historyEvents: [message: messageService.resolveCode('crossRef.package.tipps.error.title.history', null, locale),
```
hornmo: Same issue as above, put expects two parameters. Also, since the historyEvents are only defined on a title level, the correct place in the hierarchy would be `title.historyEvents`

hornmo (changes requested):

server/gokbg3/src/main/groovy/org/gokb/CrossRefPkgRun.groovy
VolkerHBZ: additionalInfo has a very complicated way of showing the Data to the User that needs further developments on the MVC view end. this is the simplified solution, until a structured way of handling it is defined. Same in the Job Overview.

VolkerHBZ: putting the errorMap in addtionalInfo didn't led to showing the Error in the (Legacy) UI

hornmo: For a structured display in the template UI, the corresponding [view](https://github.com/openlibraryenvironment/gokb/blob/g3_master/server/gokbg3/grails-app/views/apptemplates/_revreq.gsp) would need to be extended. Currently only a very specific JSON format under `additionalInfo.problems` is processed (starting on line 136), which is not compatible with the error format we are generating here.

VolkerHBZ: yeah, I know. it it exclusively used by the TSVIngestionService for one single type of errors from an unused identifier match, including suggestions on how to prevent the error. For the multitude of errors (and structures) that the import is able to generate, a proper definition of what and how to implement that view seems necessary.

hornmo:
The new test for checking the error output is still failing. It seems like the `tipp` errors are discarded for a general `global` error somehow.

hornmo (approved):

PR #424, opened 2021-02-25
3134 language

philboeselager:


server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
```diff
@@ -995,6 +999,7 @@ class BootStrap {
         RefdataCategory.lookupOrCreate('Job.Type', 'PlatformCleanup').save(flush: true, failOnError: true)
         RefdataCategory.lookupOrCreate('Job.Type', 'RecalculateStatistics').save(flush: true, failOnError: true)
 
+        LanguagesService.initialize()
```
hornmo, line 1002: Calling the LanguagesService here without setting a default URL for it in `application.yml` seems bad. Either have it not initialize at all by default and ensure some fallback values, or provide a working default.

server/gokbg3/grails-app/services/org/gokb/LanguagesService.groovy
```diff
@@ -0,0 +1,43 @@
...
+import static groovyx.net.http.Method.GET
+
+@Transactional
+class LanguagesService{
+
+  static Map localesToLanguages = [
+      "DE" : "ger",
+      "EN" : "eng",
+      "FR" : "fre"
+  ]
```
hornmo: Is this in use somewhere?

server/gokbg3/grails-app/services/org/gokb/LanguagesService.groovy
```diff
@@ -0,0 +1,43 @@
...
+
+  static Map languages = [:]
+
+  
+  static void initialize(){
+    String uriString = "${Holders.grailsApplication.config.gokb.languagesUrl}/api/listIso639two"
```
hornmo: The whole url including path should probably be the config value. Would be great to have a link to an API specification in a comment here as well.

server/gokbg3/grails-app/views/apptemplates/_tipp.gsp
```diff
@@ -62,6 +62,13 @@
                             config='KBComponent.EditStatus'/>
     </dd>
 
+    <dt>
+        <g:annotatedLabel owner="${d}" property="language">Language</g:annotatedLabel>
+    </dt>
+    <dd>
+        <g:xEditableRefData owner="${d}" field="language" config='KBComponent.Language'/>
```
hornmo: All other templates are using the category definition from KBComponent directly. Best to do it here as well for consistencies' sake.

server/gokbg3/grails-app/controllers/org/gokb/WorkflowController.groovy
```diff
@@ -1761,92 +1762,67 @@ class WorkflowController {
...
+            (tipp.title.hasProperty('editionStatement') ? sanitize(tipp.title.editionStatement) : '') + '\t' +
+            (tipp.title.hasProperty('firstEditor') ? sanitize(tipp.title.firstEditor) : '') + '\t' +
+            '\t' +  // parent_publication_title_id
+            sanitize(tipp.title?.medium?.value) + '\t' +  // publication_type
+            sanitize(tipp.paymentType?.value) + '\t' +  // access_type
+            sanitize(tipp.title.getIdentifierValue('ZDB')) +
+            '\n')
+  }
+
+
```
hornmo, line 1681: Both export functions (`packageKBartExport` as well as `packageTSVExport`) should be changed to utilize the PackageService methods like in `PackagesController.kbart()` & `PackagesController.packageTSVExport`.

server/gokbg3/grails-app/views/apptemplates/_tipp.gsp
hornmo: All other templates are extracting the category name from KBComponent directly, should be done consistently.

hornmo (changes requested):

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
philboeselager: @hornmo This was implemented likewise the `ygorUrl` example in application.yml. Do you think `ygorUrl` should be changed to a default URL as well? I thought this `null` implementation was done by intention because in reality, there is no default value for a server-specific URL. Hard-coded fallback URLs seem to be the least good solution since they are always a source of misconfiguration when changing the setup.

server/gokbg3/grails-app/services/org/gokb/LanguagesService.groovy
philboeselager: The idea is that the config file keeps the central languages API URL, whereas the single endpoints are addressed in the code. Because, unlike this initial version, further API functionality might become used. E. g. in order to show a full language-specific name of a language, you might use the endpoint `api/language` instead of holding all language names inside the (this) application. There might be further use language API use cases. You can include all endpoints in the config file, if this really your choice. The thing is just that keeping several endpoints for one microservice per `environment` crowds your config. But it's a matter of taste I would say.

server/gokbg3/grails-app/views/apptemplates/_tipp.gsp
philboeselager: I can see no inconsistency with `editStatus`. Please explain.

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
hornmo: As it stands, if no ygorUrl is provided the functionality is just not enabled at all. This on the other hand will throw an exception and stop the application from starting at all. What I meant as a fallback would be to define a small sample of RefdataValues (some commonly used languages) to be created instead if no URL is defined.

server/gokbg3/grails-app/services/org/gokb/LanguagesService.groovy
hornmo: I see your point. In that case it would be really helpful to refer to a detailed spec for each endpoint though.

server/gokbg3/grails-app/views/apptemplates/_tipp.gsp
hornmo: I meant the other templates adding `language`, where you defined the `config` value as `${org.gokb.cred.KBComponent.RD_LANGUAGE}` rather than the resolved string used in this case.

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
philboeselager: This holds for the `languagesUrl`. If it is misconfigured and pointing towards a non-existing address, `LanguagesService.initialize()` will throw an exception when being called in `Bootstrap.refdataCats()`. The same holds for its value being `null`. GOKb will simply not start then.

hornmo: Throwing an exception for an erroneous config value sounds sensible. Using the default application config though should not prevent the application from starting (especially relevant for integration testing).

philboeselager: Like we've discussed, I've changed the config value to `localhost:8070` and added the necessary documentation in the [readme](1becddb89394ef27288c51739b81f7ba51476722) and [the wiki site](https://github.com/openlibraryenvironment/gokb/wiki/GOKb-8.x-Installation-and-Bootstrapping-Guide).

server/gokbg3/grails-app/services/org/gokb/LanguagesService.groovy
philboeselager: Added documentation in G3Readme.md, the wiki and on README.md of the languages microservice.

philboeselager:
@hornmo After hours trying to find the missing links, I think it is best if you take this over. Please, let me know what has been missing here. It would be great if you can send me a note on the commits. Thank you!

hornmo:
@philboeselager ~~after some testing it looks like the property name `language` itself was the problem. After adjusting it to `primaryLanguage` instead, updating the property now works. I did not find any references to `language` being a reserved property name anywhere though.~~

Nevermind, the problem was that the property was defined in KBComponent as well as TitleInstance/TIPP, which caused the issue.

philboeselager:
> @philboeselager ~after some testing it looks like the property name `language` itself was the problem. After adjusting it to `primaryLanguage` instead, updating the property now works. I did not find any references to `language` being a reserved property name anywhere though.~
> 
> Nevermind, the problem was that the property was defined in KBComponent as well as TitleInstance/TIPP, which caused the issue.

Makes sense. I didn't see that it was still existent in both. Thanks for letting me know!

hornmo (approved):

PR #426, opened 2021-03-15
Touch component when changing historyEvent (ERMS-3221)

philboeselager:
This is my first very simple approach.
Please feel free to indicate what I might be missing.

### DIFF
--- a/server/gokbg3/grails-app/domain/org/gokb/cred/ComponentHistoryEvent.groovy
+++ b/server/gokbg3/grails-app/domain/org/gokb/cred/ComponentHistoryEvent.groovy
@@ -26,25 +26,18 @@ class ComponentHistoryEvent implements Auditable {
   }
 
 
-  def afterInsert() {
-    lastSeen = new Date().getTime()
-    touchParticipants(lastSeen)
-  }
-
-
   def afterUpdate() {
-    lastSeen = new Date().getTime()
-    touchParticipants(lastSeen)
+    touchParticipants()
   }
 
 
   def beforeDelete() {
-    lastSeen = new Date().getTime()
-    touchParticipants(lastSeen)
+    touchParticipants()
   }
 
 
-  private void touchParticipants(Long lastSeen){
+  private void touchParticipants(){
+    lastSeen = new Date().getTime()
     for (ComponentHistoryEventParticipant participant in participants){
       participant.participant.lastSeen = lastSeen
     }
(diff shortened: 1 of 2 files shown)


hornmo:
The property `lastUpdated` is reserved for auto-timestamps by GORM and cannot be manually set (see http://gorm.grails.org/6.1.x/hibernate/manual/#eventsAutoTimestamping). Alternatives would be `lastSeen` (as a `Long` via `Date.getTime()` ) or a descriptive `lastUpdatedComment` `String`.

hornmo (changes requested):
See previous comment

philboeselager:
> The property `lastUpdated` is reserved for auto-timestamps by GORM and cannot be manually set (see http://gorm.grails.org/6.1.x/hibernate/manual/#eventsAutoTimestamping). Alternatives would be `lastSeen` (as a `Long` via `Date.getTime()` ) or a descriptive `lastUpdatedComment` `String`.

@hornmo I'm preferring `lastSeen` here, since changing `lastUpdatedComment` might conflict with other use cases of this field. Using `lastSeen` will lead to some unnecessary updates, right? But this should be OK, as we should be able to handle the overall number.

server/gokbg3/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -471,7 +471,9 @@ class FTUpdateService {
 
       def total = 0
       Date from = new Date(latest_ft_record.lastTimestamp)
-      def countq = domain.executeQuery("select count(o.id) from " + domain.name + " as o where (( o.lastUpdated > :ts ) OR ( o.dateCreated > :ts )) ", [ts: from], [readonly: true])[0]
+      Long seen = latest_ft_record.lastSeen
+      def countq = domain.executeQuery("select count(o.id) from " + domain.name + " as o where ((o.lastUpdated > :ts) "+
+          "OR (o.dateCreated > :ts) OR (o.lastSeen > :seen)) ", [ts: from, seen: seen], [readonly: true])[0]
```
hornmo: Since the change in `lastSeen` triggers the auto timestamp and adjusts `lastUpdated`, the changes here are not necessary.

server/gokbg3/grails-app/domain/org/gokb/cred/ComponentHistoryEvent.groovy
```diff
@@ -9,17 +9,44 @@ class ComponentHistoryEvent implements Auditable {
...
+
   String getLogEntityId() {
       "${this.class.name}:${id}"
   }
 
+
+  def afterInsert() {
+    lastSeen = new Date().getTime()
+    touchParticipants(lastSeen)
+  }
```
hornmo: This might be a bit more complicated, since the event has to be created before its participants. This means we would probably only need the `afterUpdate` (since adding/removing a participant from `participants` should trigger an update) and should check if two participants exist before doing the touch.

hornmo (changes requested):

hornmo (approved):

PR #447, opened 2021-06-11
copying title data (name, identifiers) to empty tipp fields

VolkerHBZ:


hornmo:
Tried this on a set of about 40k TIPPs and it looks like there is a memory leak (at least in interactive mode).

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
```diff
@@ -0,0 +1,77 @@
...
+              count++
+            }
+            tipp.save(flush:true)
+            tipp.finalize()
+          }
+          job?.setProgress(index, tippIDs.size())
+          if (job?.isCancelled()) {
+            cancelled = true
+          }
+          if (count > 100) {
```
hornmo: Ensure this is consistently called, either by incrementing `count` for each TIPP or using smth. like `index % 100 == 0`

hornmo (changes requested):

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
VolkerHBZ: The variable count gets incremented each time a TIPP gets changed, after the session.flush() count is reset to zero and gets incremented again... this seemed consistent enough to me.

Adopted your proposition.

hornmo: The problem I was running in to was that the count is not increased if a TIPP is skipped or nothing is changed. If there are a lot of TIPPs that already have all Identifiers attached to them, those will not increase the counter currently and cause the session to balloon until enough TIPPs are actually changed.

VolkerHBZ: I noted that too and changed the implementation to: if (index%100==0)

hornmo (approved):

PR #448, opened 2021-06-11
Cache OAI package records

hornmo:
- saves & loads XML(ish) files in tmp
- currently only triggered by an OAI request for the specific item


VolkerHBZ (approved):
if an export fails, the unfinished XML-file in the tmp directory might make subsequent imports impossible.

VolkerHBZ (approved):

PR #453, opened 2021-08-10
Allocate ReviewRequest to CuratoryGroup (not User) (GOKbIT-83)

philboeselager:


hornmo:
Rather than changing `allocatedTo`, I would prefer doing this via the `AllocatedReviewGroup` linking object that is currently used for representing CuratoryGroup allocation.

philboeselager:
> Rather than changing `allocatedTo`, I would prefer doing this via the `AllocatedReviewGroup` linking object that is currently used for representing CuratoryGroup allocation.

... and remove `allocatedTo` (User) then? 

But how do we determine which CG is the one currently being "assigned" to that RR then, @hornmo ?

hornmo:
> ... and remove allocatedTo (User) then?

Might as well, the current workflows don't set that value in any case. Just cutting it from the domain class will not be sufficient though since it is still referenced in various places, most prominently ReviewRequestService.raise() through which all RRs are preferably created (the `raise` in the domain class should probably be deprecated).

> But how do we determine which CG is the one currently being "assigned" to that RR then, @hornmo ?

An option would be to extend/replace the values of `AllocatedReviewGroup.status` (currently the default values `Claimed` and `In Progress` are defined in Bootstrap), which is unused as of now. Maybe something like `Initiator` + `Elevated`?
If we want to filter by these criteria in the frontend, we would need to add this to the custom lookup filter in `ComponentLookupService.restLookup`:

https://github.com/openlibraryenvironment/gokb/blob/ea98d8536042eecda187c3f19de7bb6dfc6449af/server/gokbg3/grails-app/services/org/gokb/ComponentLookupService.groovy#L457-L484

philboeselager:
> An option would be to extend/replace the values of `AllocatedReviewGroup.status` (currently the default values `Claimed` and `In Progress` are defined in Bootstrap), which is unused as of now. Maybe something like `Initiator` + `Elevated`?

Not sure if we can reflect transferring RRs back to the initiator CG using this approach. Can we, @hornmo ?

Technically, we could implement that by introducing boolean attribues like `initiative` and `assigned` (active) to ARG. This would potentially end up in having more than 1 ARG connecting a specific RR to a specific CG. Not sure if we would want that.

Maybe, we can implement it with ARG changing the member variable `group` to `groups`, implementing `groups` as an (time-) ordered list. The first entry is always the initiator then. The "assigned" CG could either be pointed to by an index (being "0" when pointing to the initiator, or it could be like `CuratoryGroup assigned = groups[0]`. (Just for sketching the idea.)

hornmo:
> Technically, we could implement that by introducing boolean attribues like initiative and assigned (active) to ARG. This would potentially end up in having more than 1 ARG connecting a specific RR to a specific CG. Not sure if we would want that.

We may also add a direct link between the ARGs themselves via smth like escalatedFrom when we create the new ARG?

philboeselager:
> We may also add a direct link between the ARGs themselves via smth like escalatedFrom when we create the new ARG?

Yes, thought about that, too. This would mean we set the status of the initiator to "deactivated", right?

hornmo:
>Yes, thought about that, too. This would mean we set the status of the initiator to "deactivated", right?

Yeah. either a generic `Inactive` or more specific `Escalated`, both would be okay I think.

philboeselager:
@hornmo I've implemented the outline of this idea in https://github.com/openlibraryenvironment/gokb/pull/454 . This is just the functionality itself, no use cases of it, yet. Let's further discuss there.

PR #454, opened 2021-08-11
Add escalation routine for AllocatedReviewGroup (GOKbIT-83)

philboeselager:


server/gokbg3/grails-app/controllers/org/gokb/rest/CuratoryGroupsController.groovy
```diff
@@ -117,7 +119,8 @@ class CuratoryGroupsController {
     def errors = [:]
 
     if (group) {
-      def qry = "where exists (select arg from AllocatedReviewGroup arg where arg.group = :group and arg.review = rr)"
+      def qry = "where exists (select arg from AllocatedReviewGroup arg" +
+                              "where arg.group = :group and arg.review = rr and arc.status != :inactive)"
```
hornmo, line 126: The query param `:inactive` has not been defined here. Need to retrieve the RDV.

server/gokbg3/grails-app/domain/org/gokb/cred/AllocatedReviewGroup.groovy
```diff
@@ -7,12 +7,13 @@ class AllocatedReviewGroup implements Serializable {
   CuratoryGroup group
   ReviewRequest review
   RefdataValue status
+	AllocatedReviewGroup escalatedFrom
```
hornmo, line 10: We will have to handle the case where the escalating ARG defined here is deleted. We may want to delete this ARG as well in that case.

server/gokbg3/grails-app/controllers/org/gokb/GroupController.groovy
```diff
@@ -46,14 +46,14 @@ class GroupController {
...
-   
+
       log.debug("Got ${cg_components.size()} connected components")
 
       def cg_review_tasks_hql = ''' from ReviewRequest as rr where ((
         rr.allocatedTo in ( select u from CuratoryGroup as cg join cg.users as u where cg = :group )
         or rr.componentToReview.id in (:cgcomponents)
       ) or exists (select arc from AllocatedReviewGroup as arc where arc.review = rr and arc.group = :group))
-      and rr.status != :closed and rr.status != :deleted
+      and rr.status != :closed and rr.status != :deleted and arc.status != :inactive
```
hornmo: The RefdataValue has to be retrieved for `:inactive`.

server/gokbg3/grails-app/services/org/gokb/ComponentLookupService.groovy
```diff
@@ -478,7 +477,7 @@ class ComponentLookupService {
           hqlQry += " AND "
         }
 
-        hqlQry += "exists (select 1 from AllocatedReviewGroup as ag where ag.review = p and ag.group IN :alg)"
+        hqlQry += "exists (select 1 from AllocatedReviewGroup as ag where ag.review = p and ag.group IN :alg and ag.status != :inactive)"
```
hornmo, line 537: See above

hornmo (changes requested):

server/gokbg3/grails-app/domain/org/gokb/cred/AllocatedReviewGroup.groovy
philboeselager: There is : 

- https://github.com/openlibraryenvironment/gokb/blob/f767a613ffa8a3b500deea2f3e626e54e6f4634e/server/gokbg3/grails-app/domain/org/gokb/cred/AllocatedReviewGroup.groovy#L46
- https://github.com/openlibraryenvironment/gokb/blob/f767a613ffa8a3b500deea2f3e626e54e6f4634e/server/gokbg3/grails-app/domain/org/gokb/cred/AllocatedReviewGroup.groovy#L50
- https://github.com/openlibraryenvironment/gokb/blob/f767a613ffa8a3b500deea2f3e626e54e6f4634e/server/gokbg3/grails-app/domain/org/gokb/cred/KBComponent.groovy#L1360
- https://github.com/openlibraryenvironment/gokb/blob/f767a613ffa8a3b500deea2f3e626e54e6f4634e/server/gokbg3/grails-app/domain/org/gokb/cred/KBComponent.groovy#L1397

If I understand correctly, in all of these cases the `escalatedFrom` ARG is deleted anyways.

philboeselager:
@hornmo Added the parameters (see a7acf32). Still, e. g. calling `group/index/<id>` is not working.

Exception is: `org.hibernate.hql.internal.ast.QuerySyntaxException: Invalid path: 'arc.status_id'`

I could not yet find out what the code problem is. Can you?

hornmo:
>@hornmo Added the parameters (see a7acf32). Still, e. g. calling group/index/<id> is not working.
Exception is: org.hibernate.hql.internal.ast.QuerySyntaxException: Invalid path: 'arc.status_id'
I could not yet find out what the code problem is. Can you?

For HQL Queries you need to use the domain class property name (so `arc.status`) rather than the database row `status_id`.
Also the status check has to go into the subquery where arc is defined, after `select arc from AllocatedReviewGroup as arc where arc.review = rr and arc.group = :group`.

philboeselager:
@hornmo If tested fine, the backend might generally be finished. We might be wanting to add a `reviews/deescalatable` check - see https://github.com/openlibraryenvironment/gokb-ui/pull/55#issuecomment-921004992 .

@darupp I think we lost track of the discussion while trying to define the names of the CG categories. Please send them to hornmo so that he can add them to the Bootstrap.

server/gokbg3/grails-app/domain/org/gokb/cred/CuratoryGroupType.groovy
```diff
@@ -0,0 +1,40 @@
+package org.gokb.cred
+
+class CuratoryGroupType extends KBComponent{
```
hornmo: No need to extend from KBComponent here.

server/gokbg3/grails-app/services/org/gokb/ReviewRequestService.groovy
```diff
@@ -57,4 +56,15 @@ class ReviewRequestService {
     }
     req
   }
+
+
+  AllocatedReviewGroup escalate(AllocatedReviewGroup arg, CuratoryGroup cg){
+    arg.status = RefdataCategory.lookup('AllocatedReviewGroup.Status', 'Inactive')
+    AllocatedReviewGroup result = AllocatedReviewGroup.create(cg, arg.review, false)
```
hornmo: We should always check for a duplicate beforehand.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -387,7 +452,82 @@ class ReviewsController {
       linksObj.update = [href:base]
       linksObj.delete = [href:base]
     }
-
     return linksObj
   }
+
+
+  private JSON getEscalationTargetGroupId(def rrId, def activeGroupId, def params){
```
hornmo: This returns a Map and is only converted to JSON at rendering.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -355,16 +353,83 @@ class ReviewsController {
...
+  def isEscalatable() {
+    render getEscalationTargetGroupId(params.id, params.activeGroupId, params) as JSON
+  }
+
+
+  @Secured(value=["hasRole('ROLE_EDITOR')", 'IS_AUTHENTICATED_FULLY'])
+  @Transactional
+  def escalate(){
+    def result = getEscalationTargetGroupId(params.id, request.JSON?.activeGroup?.id, params)
+    if (result.target?.result == 'OK'){
```
hornmo: Check should be for `result.isEscalatable`.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -355,16 +353,83 @@ class ReviewsController {
...
+      CuratoryGroup escalatingGroup = CuratoryGroup.findById(result.target.escalatingGroup)
```
hornmo: Now should be just `result.escalatingGroup`.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -355,16 +353,83 @@ class ReviewsController {
...
+      CuratoryGroup targetGroup = escalatingGroup.superordinatedGroup
```
hornmo: The target group should already be determined by `getEscalationTargetGroupId`.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -355,16 +353,83 @@ class ReviewsController {
...
+      AllocatedReviewGroup arg = AllocatedReviewGroup.findByGroup(escalatingGroup)
```
hornmo: This needs to specify the review as well.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -355,16 +353,83 @@ class ReviewsController {
...
+      AllocatedReviewGroup newArg = reviewRequestService.escalate(arg, targetGroup)
+      if (newArg){
+        ReviewRequest rr = ReviewRequest.get(genericOIDService.oidToId(params.id))
+        def inactive = RefdataCategory.lookup('AllocatedReviewGroup.Status', 'Inactive')
+        rr.allocatedGroups.each{ ag -> ag.status = inactive }
+        newArg.status = RefdataCategory.lookup('AllocatedReviewGroup.Status', 'In Progress')
+        result.target.message = "The requested ReviewRequest has been escalated."
+      }
```
hornmo: Status changes are already handled by the service method.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -387,7 +452,82 @@ class ReviewsController {
...
+      result.message = "ReviewRequest for id ${rrId} may not be edited."
+      return result
+    }
+
+    CuratoryGroup escalatingGroup = null
+    if (activeGroupId){
+      String id = String.valueOf(activeGroupId)
+      if (!StringUtils.isEmpty(id)){
+        CuratoryGroup toBeChecked = CuratoryGroup.findById(id)
+        List<AllocatedReviewGroup> argCandidates = AllocatedReviewGroup.findAllByGroupAndReview(toBeChecked, rr)
```
hornmo: We may even filter by status (`In Progress`) as well, to keep out already escalated ARGs.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -387,7 +452,82 @@ class ReviewsController {
...
+        return result
+      }
+      if (inProgressARGs.size() > 1){
+        response.setStatus(409)
+        result.message = "The requested ReviewRequest can not be escalated due to multiple group allocations being in progress."
+        return result
+      }
+      escalatingGroup = inProgressARGs.get(0).group
+      result.escalatingAllocatedReviewGroup = inProgressARGs.get(0).id
+    }
```
hornmo: We can skip this. If the CG has no ARG for this review, we just throw an error.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -387,7 +452,82 @@ class ReviewsController {
...
+
+    if (!escalatingGroup){
+      response.setStatus(404)
+      result.message = "Could not get curatory group to be escalated from."
+      return result
+    }
+
+    CuratoryGroup escalatedToCG = escalatingGroup.superordinatedGroup
```
hornmo, line 559: This is only a very specific case and not generalizable:

- Only Reviews with specific types of `componentToReview` may be escalated (currently only `TitleInstance` .
- CGs _without_ a superordinate Group (so they should have no `type` set (?)) OR groups marked as "Title"-CGs may currently _only_ escalate reviews with a `JournalInstance` as `componentToReview`. These are escalated directly to the group defined currently by the deployment config value `grailsApplication.config.gokb.editorialAdmin?.journals`  (in currently deployed instances 'ZDB').

hornmo (changes requested):

server/gokbg3/grails-app/domain/org/gokb/cred/CuratoryGroupType.groovy
hornmo: This breaks in Bootstrap and therefore prevents running the App.

hornmo (changes requested):

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: According to your last message, this is now extended to `BookInstance`, `DatabaseInstance` and `JournalInstance`, right?

server/gokbg3/grails-app/domain/org/gokb/cred/CuratoryGroupType.groovy
philboeselager: Breaks without `extends KBComponent` in my local GOKb instance. Should be a matter of internal DB structure.

server/gokbg3/grails-app/services/org/gokb/ReviewRequestService.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Fixed.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -361,12 +362,95 @@ class ReviewsController {
...
+
+  @Secured(value=["hasRole('ROLE_EDITOR')", 'IS_AUTHENTICATED_FULLY'])
+  @Transactional
+  def escalate(){
+    def result = getEscalationTargetGroupId(params.id, request.JSON?.activeGroup, params)
+    if (result.isEscalatable){
+      CuratoryGroup escalatingGroup = CuratoryGroup.findById(result.escalatingGroup)
+      CuratoryGroup targetGroup = escalatingGroup.superordinatedGroup
+      AllocatedReviewGroup arg = AllocatedReviewGroup.findByGroupAndReview(escalatingGroup, ReviewRequest.findById(params.id))
+      AllocatedReviewGroup newArg = reviewRequestService.escalate(arg, targetGroup)
```
hornmo: Target should be the group from `result.escalationTargetGroup`.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -395,10 +479,22 @@ class ReviewsController {
...
+    if (obj.allocatedTo == user) {
+      curator = true
+    }
+    else if (obj.allocatedGroups?.group.id.intersect(user.curatoryGroups?.id)) {
+      curator = true
+    }
+    curator
+  }
+
+
```
hornmo: There is a generic method `componentUpdateService.isUserCurator(obj, user)` which should be used instead of creating local methods.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -408,7 +504,93 @@ class ReviewsController {
...
+    if (!rr.componentToReview?.class in [BookInstance.class, DatabaseInstance.class, JournalInstance.class]){
+      response.setStatus(405)
+      result.message = "ReviewRequest belongs to the un-escalatable class ${rr.componentToReview?.class().getSimpleName()}"
+      return result
+    }
+
+    CuratoryGroup escalatingGroup = null
+    if (activeGroupId){
+      String id = String.valueOf(activeGroupId)
+      if (!StringUtils.isEmpty(id)){
```
hornmo: Why not require the ID to be `int` from the start?

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
```diff
@@ -1021,20 +982,66 @@ class BootStrap {
...
+            new CuratoryGroupType(level: CuratoryGroupType.Level.PACKAGE, name: "Journal Package Curators").save(flush: true, failOnError: true)
+        CuratoryGroupType.findByName("E-Book Package Curators") ?:
+            new CuratoryGroupType(level: CuratoryGroupType.Level.PACKAGE, name: "E-Book Package Curators").save(flush: true, failOnError: true)
+        CuratoryGroupType.findByName("Journal Title Curators") ?:
+            new CuratoryGroupType(level: CuratoryGroupType.Level.TITLE, name: "Journal Title Curators").save(flush: true, failOnError: true)
+        CuratoryGroupType.findByName("E-Book Title Curators") ?:
+            new CuratoryGroupType(level: CuratoryGroupType.Level.TITLE, name: "E-Book Title Curators").save(flush: true, failOnError: true)
+        CuratoryGroupType.findByName("Journal Central Curators") ?:
+            new CuratoryGroupType(level: CuratoryGroupType.Level.CENTRAL, name: "Journal Central Curators").save(flush: true, failOnError: true)
+    }
```
hornmo, line 1039: What about groups that do not distinguish between the different publication types on one of these levels (PACKAGE/TITLE) or both? This solution would mean that for each institution the differentiation always has to be the same for both levels. This would probably entail some users having to belong to two arbitrarily split groups, which would not be desirable.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -408,7 +504,93 @@ class ReviewsController {
...
+    }
+
+    CuratoryGroup editorialGroup = grailsApplication.config.gokb.editorialAdmin?.journals ? CuratoryGroup.findByNameIlike(grailsApplication.config.gokb.editorialAdmin.journals) : null
+    CuratoryGroup escalatedToCG = escalatingGroup.superordinatedGroup
+
+    if (!escalatedToCG){
+      response.setStatus(409)
+      result.message = "Could not escalate due to missing superordinated group."
+      return result
+    }
```
hornmo, line 567: Only give up here if there is no `editorialGroup` for the respective class of componentToReview.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Yes. Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Yes. I actually used the `componentUpdateService` but forgot to remove the local method. Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
philboeselager: That's a very good question. To my believe, the answer will not be given within this ticket.
Put simple, we can say that there is no need for escalation if a user is responsible for both, Package and Title level.
The good thing is that the escalation level is only of declarative function. Still, any user can solve any RR that is assigned to his CG.
@hornmo If you agree about this question being solved in a later ticket, please resolve the conversation here.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: ... both are exactly the same, but yes, this is more stringent.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: ... both are exactly the same, but yes, this is more stringent.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Now, this is directly referenced.

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
philboeselager: There is another Ticket, GOKbIT-118, which handles this issue.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -408,7 +497,100 @@ class ReviewsController {
...
+
+    if (!escalatingGroup){
+      response.setStatus(404)
+      result.message = "Could not get curatory group to be escalated from."
+      return result
+    }
+
+    CuratoryGroup editorialGroup = grailsApplication.config.gokb.centralGroups && grailsApplication.config.gokb.centralGroups['componentClass'] ?
+        CuratoryGroup.findByNameIlike(grailsApplication.config.gokb.centralGroups['componentClass']) : null
```
hornmo: Should be `.centralGroups[componentClass]` here instead of using the String in the lookup.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -408,7 +497,100 @@ class ReviewsController {
...
+        RefdataValue inProgress = RefdataCategory.lookup('AllocatedReviewGroup.Status', 'In Progress')
+        List<AllocatedReviewGroup> argCandidates = AllocatedReviewGroup.findAllByGroupAndReviewAndStatus(toBeChecked, rr, inProgress)
+        if (argCandidates.size() == 1){
+          escalatingGroup = toBeChecked
+        }
+      }
+    }
+
+    if (!escalatingGroup){
+      response.setStatus(404)
```
hornmo: This does not neccessarily mean 404, since there may be an ARG that is already elevated (= Inactive & not in argCandidates) and that should not be considered an error.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -408,7 +497,100 @@ class ReviewsController {
...
+    if (!rr){
+      response.setStatus(404)
+      result.message = "ReviewRequest not found for id ${rrId}."
+      return result
+    }
+    if (!rr.isEditable()){
+      response.setStatus(403)
+      result.message = "ReviewRequest for id ${rrId} may not be edited."
+      return result
+    }
```
hornmo, line 517: We might put this in the request context instead, since this is only really looking up the ID and not checking user permissions.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -408,7 +497,100 @@ class ReviewsController {
...
+    if (!deescalatingGroup){
+      response.setStatus(404)
+      result.message = "Could not get curatory group to be deescalated from."
+      return result
+    }
+    AllocatedReviewGroup deescArg, targetArg
+    deescArg = AllocatedReviewGroup.findByGroupAndReview(deescalatingGroup, rr)
+    targetArg = deescArg?.escalatedFrom ?: null
+    if (deescArg && targetArg){
+      response.setStatus(200)
```
hornmo: 200 is standard so no need to set it explicitly

hornmo (changes requested):
Found some issues testing with the latest frontend changes. We might also want additional tests in `server/gokbg3/src/integration-test/groovy/gokbg3/rest/CuratoryGroupsTestSpec.groovy ` to do some checks.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
hornmo: Also, in the following check we should exclude `escalatingGroup == editorialGroup`.

philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: This is a private method. Permissions are checked in the two calling methods, `escalate()` and `isEscalatable()`. Should be good like this. Otherwise, please specify and reopen conversation, @hornmo .

server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
philboeselager: Done.

server/gokbg3/grails-app/services/org/gokb/ReviewRequestService.groovy
```diff
@@ -57,4 +56,22 @@ class ReviewRequestService {
...
+  AllocatedReviewGroup escalate(AllocatedReviewGroup arg, CuratoryGroup cg){
+    arg.status = RefdataCategory.lookup('AllocatedReviewGroup.Status', 'Inactive')
+    AllocatedReviewGroup result = AllocatedReviewGroup.findByGroupAndReview(cg, arg.review)
+    if (result == null){
+      result = AllocatedReviewGroup.create(cg, arg.review, true)
+    }
+    arg.review.allocatedGroups.removeAll()
+    arg.review.allocatedGroups << cg
+    arg.review.save()
+    arg.save()
```
hornmo: This has been revised to be replaced by:
`AllocatedReviewGroup result = AllocatedReviewGroup.findByGroupAndReview(cg, arg.review) ? AllocatedReviewGroup.create(cg, arg.review, false)`

hornmo (changes requested):

server/gokbg3/grails-app/services/org/gokb/ReviewRequestService.groovy
philboeselager: Done.

server/gokbg3/grails-app/domain/org/gokb/cred/CuratoryGroupType.groovy
philboeselager: Fixed.

hornmo (approved):

PR #459, opened 2021-08-18
added support for prices

VolkerHBZ:
in the REST interface by extending TitleInstancePackagePlatform.upsertDTO


hornmo (commented):
`TippController.update` did not use `TIPP.upsertDTO` up until now, but this should work now.

hornmo (approved):

PR #462, opened 2021-08-25
Es label urls

hornmo:
- Fix ES label query
- Generalize Refdata Queries
- Add missing Refdata mappings

### DIFF
--- a/server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
+++ b/server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
@@ -671,6 +671,10 @@ class ESSearchService{
             sortBy = "sortname"
           }
 
+          if (sortBy == "lastUpdated") {
+            sortBy = "lastUpdatedDisplay"
+          }
+
           if (ESWrapperService.mapping.component.properties[sortBy]?.type == 'text') {
             errors['sort'] = "Unable to sort by text field ${sortBy}!"
           }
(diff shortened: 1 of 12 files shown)


server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
```diff
@@ -318,33 +322,29 @@ class ESSearchService{
...
+  private void addRefdataToQuery(String value, QueryBuilder refdataQuery, String field){
     try{
-      status = RefdataValue.get(Long.valueOf(status))
+      value = RefdataValue.get(Long.valueOf(value))?.value
     }
     catch (Exception e){
     }
-    statusQuery.should(QueryBuilders.termQuery('status', status))
+    refdataQuery.should(QueryBuilders.matchQuery(field, value))
   }
```
VolkerHBZ, line 348: this is quite a wild construct - who invented it?

VolkerHBZ (approved):

server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
hornmo: Do you mean the whole method or the line?

VolkerHBZ: the whole method: parsing the string for numbers and ignoring the possible exception while using the (un-)altered value afterwards... Clearly, it's working, but looks... unusual and therefore weird.

hornmo: I agree it looks weird, but allows us to handle both id and value requests in a simple manner. I did not write this specific piece but I'm also fine with it as it is.

PR #464, opened 2021-08-27
Coverage matching 112

VolkerHBZ:


server/gokbg3/grails-app/services/org/gokb/TippService.groovy
```diff
@@ -272,6 +306,38 @@ class TippService {
...
+    }
+  }
+
+  private TIPPCoverageStatement latest(def covStmts) {
+    def latest = null
+    if (covStmts?.size() > 0) {
+      def today = LocalDate.now()
+      covStmts.each {
+        if ((latest == null || latest.startDate < it.startDate)
+            && (today.isAfter(it.startDate.toInstant().atZone(ZoneId.systemDefault()).toLocalDate())))
```
hornmo: This will fail if startDate is null, so we have to filter for that.

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
```diff
@@ -272,6 +306,38 @@ class TippService {
...
+  private void coverageCheck(tipp, found) {
+    // find the latest coverage
+    TIPPCoverageStatement latest = latest(tipp.coverageStatements)
+    if (latest && found.matches.size > 1) {
+      // too many identifier matches
+      def covMatch = []
+      found.matches.each { comp ->
+        if (JournalInstance.isInstance(comp)) {
+          JournalInstance journal = JournalInstance(comp)
+          if (journal.publishedFrom == latest.startDate && journal.publishedTo == latest.endDate) {
```
hornmo: We have to compare the ranges here instead, so something like `startDate >= publishedFrom && endDate <= publishedTo`, complicated a bit since each of the values is optional.

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
```diff
@@ -177,11 +202,17 @@ class TippService {
           'medium', 'language'
       ], tipp, ti)
 
+      def pubFrom = tipp.accessStartDate ? GOKbTextUtils.completeDateString(tipp.accessStartDate.format("yyyy-MM-dd")) : null
+      def pubTo = tipp.accessEndDate ? GOKbTextUtils.completeDateString(tipp.accessEndDate.format("yyyy-MM-dd"), false) : null
```
hornmo: We cannot derive the pubFrom / pubTo from access dates. If anything we could check a coverageStatement for `startVolume 1 && startDate`, but even that sounds a bit shaky. Better to leave this to the ZDB Augment job.

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
```diff
@@ -177,11 +202,17 @@ class TippService {
...
       def firstInPrint = tipp.dateFirstInPrint ? GOKbTextUtils.completeDateString(tipp.dateFirstInPrint.format("yyyy-MM-dd")) : null
       def firstOnline = tipp.dateFirstOnline ? GOKbTextUtils.completeDateString(tipp.dateFirstOnline.format("yyyy-MM-dd")) : null
 
-      title_changed |= ti.hasProperty('dateFirstInPrint')?ClassUtils.setDateIfPresent(firstInPrint, ti, 'dateFirstInPrint'):false
-      title_changed |= ti.hasProperty('dateFirstOnline')?ClassUtils.setDateIfPresent(firstOnline, ti, 'dateFirstOnline'):false
+      log.debug("Completed date publishedFrom ${tipp.accessStartDate} -> ${pubFrom}")
+
+      title_changed |= ti.hasProperty('publishedFrom') ? ClassUtils.updateDateField(pubFrom, ti, 'publishedFrom') : false
+      title_changed |= ti.hasProperty('publishedTo') ? ClassUtils.updateDateField(pubTo, ti, 'publishedTo') : false
```
hornmo: See above.

hornmo (changes requested):

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
VolkerHBZ: changed

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
VolkerHBZ: adapted it

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
VolkerHBZ: deleted

server/gokbg3/grails-app/services/org/gokb/TippService.groovy
VolkerHBZ: dito

hornmo (approved):

PR #469, opened 2021-10-25
Adjust review generation for TIPP import

hornmo:
Draft for new TIPP import review types, pending UI updates ...


philboeselager:
Looks good to me.

hornmo:
Now includes fixes & changes for tickets (145, 195 & 196)

server/gokbg3/src/main/groovy/org/gokb/UpdatePkgTippsRun.groovy
```diff
@@ -705,50 +836,40 @@ class UpdatePkgTippsRun {
...
+
+      //     if (tipp) {
+      //       tipps << tipp
+      //     }
+      //     else  {
+      //       log.warn("ES record TIPP ${it.uuid} does not exist!")
+      //     }
+      //   }
+      //   return tipps
+      // }
```
philboeselager, line 882: Maybe we can add a comment on how long we are wanting to keep out-commented code in the future. Sometimes, these blocks get really old and every time you pass by you ask yourself: Should I delete that?

philboeselager (approved):

server/gokbg3/src/main/groovy/org/gokb/UpdatePkgTippsRun.groovy
hornmo: Great Idea, I added comments with version info.

PR #473, opened 2022-01-06
139 ezb enrichment

philboeselager:


hornmo (commented):
This raises some questions we have to discuss, mainly if we even want to create ReviewRequests for EZB conflicts.

The changes in TippService seem out of place and don't really make sense to me.

server/gokbg3/grails-app/services/org/gokb/TitleAugmentService.groovy
```diff
@@ -203,12 +205,106 @@ class TitleAugmentService {
     }
   }
 
+  def augmentEzb(titleInstance) {
+    log.debug("Augment EZB - TitleInstance: ${titleInstance.niceName} - ${titleInstance.class?.name}")
+    CuratoryGroup editorialGroup = grailsApplication.config.gokb.ezbAugment?.rrCurators ?
+        CuratoryGroup.findByNameIlike(grailsApplication.config.gokb.ezbAugment.rrCurators) : null
```
hornmo: if augment is activated via config, we should probably create the `CuratoryGroup` here if it doesn't exist.

server/gokbg3/grails-app/init/gokbg3/BootStrap.groovy
```diff
@@ -235,7 +236,9 @@ class BootStrap {
         log.debug("Checking for missing component statistics")
         ComponentStatisticService.updateCompStats()
 
-        log.info("GoKB Init complete");
+        AugmentJob.schedule(grailsApplication.config.gokb.augment.cron)
```
hornmo: This does not seem to work right now, and I cannot find the corresponding method.

server/gokbg3/grails-app/services/org/gokb/TitleAugmentService.groovy
```diff
@@ -203,12 +205,106 @@ class TitleAugmentService {
...
+      }
+      else {
+        log.debug("Skipping title with existing RR ..")
+      }
+    }
+  }
+
   private void setNewTitleInfo(titleInstance, info) {
-    if (!titleInstance.publishedFrom && info.publishedFrom) {
+    if ((!titleInstance.hasProperty('publishedFrom') || !titleInstance.publishedFrom) && info.publishedFrom) {
```
hornmo: The property is defined in the generic base class `TitleInstance`, therefore this check shouldn't be necessary.

hornmo (changes requested):
The changes in `TippService` are overhauled and should be removed from this PR.

philboeselager (commented):
can't be approved by the raiser of the PR. Feel dree, @hornmo :-)

hornmo (approved):

PR #477, opened 2022-01-25
fix package combo update

hornmo:
- Fix checks for package combo updates
- Fix duplicate cg link assignment
- Fix Platform matching name lookup

### DIFF
--- a/server/gokbg3/grails-app/controllers/org/gokb/rest/CuratoryGroupsController.groovy
+++ b/server/gokbg3/grails-app/controllers/org/gokb/rest/CuratoryGroupsController.groovy
@@ -287,6 +287,7 @@ class CuratoryGroupsController {
       }
       else{
         try{
+          superordinate.subordinatedGroups << subordinate
           subordinate.superordinatedGroup = superordinate
           result.result = 'OK'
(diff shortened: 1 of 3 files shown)


server/gokbg3/grails-app/controllers/org/gokb/rest/CuratoryGroupsController.groovy
```diff
@@ -287,7 +287,6 @@ class CuratoryGroupsController {
       }
       else{
         try{
-          superordinate.subordinatedGroups << subordinate
```
philboeselager: I think we need this for "re-escalation" (a superordinated CG giving the issue back to the subordinated CG) . Don't we?

philboeselager (changes requested):

server/gokbg3/grails-app/controllers/org/gokb/rest/CuratoryGroupsController.groovy
hornmo: I moved this out of here since this is an unrelated issue.

philboeselager (approved):

PR #478, opened 2022-01-25
fix curatoryGroup domain mapping & add more checks

hornmo:
- Added more error states in CG connection
- Add explicit GORM mapping for the new hasMany relationship between CuratoryGroups
- Remove duplicate assignment & add save

### DIFF
--- a/server/gokbg3/grails-app/controllers/org/gokb/rest/CuratoryGroupsController.groovy
+++ b/server/gokbg3/grails-app/controllers/org/gokb/rest/CuratoryGroupsController.groovy
@@ -281,6 +281,7 @@ class CuratoryGroupsController {
     else {
       CuratoryGroup superordinate = CuratoryGroup.get(genericOIDService.oidToId(params.superordinateId))
       CuratoryGroup subordinate = CuratoryGroup.get(genericOIDService.oidToId(params.subordinateId))
+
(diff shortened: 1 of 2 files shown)


server/gokbg3/grails-app/controllers/org/gokb/rest/CuratoryGroupsController.groovy
```diff
@@ -271,30 +271,43 @@ class CuratoryGroupsController {
...
+        response.status = 404
         result.message = "CuratoryGroup combination not found for superordinate id ${params.superordinateId} and subordinate id ${params.subordinateId}."
       }
-      else if (!(superordinate.type.level > subordinate.type.level)){
-        response.setStatus(409)
+      else if (!subordinate.type || !superordinate.type) {
+        response.status = 400
+        result.message = "One or both of the provided groups do not have any type set."
+      }
+      else if (superordinate.type.level.levelNumber == 75) {
```
philboeselager: `CuratoryGroupType.Level.CENTRAL.value` is more typing than `75` but still preferable.

philboeselager (changes requested):
Looks good overall. Only 1 comment.

philboeselager (approved):

PR #480, opened 2022-02-10
124 rest price updates

hornmo:


server/gokbg3/grails-app/services/gokbg3/RestMappingService.groovy
```diff
@@ -151,8 +152,24 @@ class RestMappingService {
         }
         else {
           switch (p.type) {
+            case Float.class:
+              if (p.name == 'price') {
+                String pstring = obj[p.name] ? "${obj[p.name].trunc(2)}" : null
+
+                if (pstring) {
+                  if (!pstring.contains('.')) {
```
philboeselager, line 160: Are we sure that pstring does not contain other separator chars, esp. `,` ?

philboeselager (commented):
1 comment, rest is fine.

server/gokbg3/grails-app/services/gokbg3/RestMappingService.groovy
hornmo: I would think so, `Float.toString()` should always be of the form `123.23345`

philboeselager: Sure. I missed that.

philboeselager (approved):

PR #481, opened 2022-02-10
Don't set accessStartDate to package date (GOKbIT-225)

philboeselager:


hornmo (changes requested):
```
Erwartetes Ergebnis:

    Dieser Wert entspricht dem Datum. das im KBART-Dateinamen als letzter Parameter angegeben ist, da keine Spalte "access_start_date" vorhanden ist: Springer_Global_Springer_Wirtschaftswissenschaften_eBooks_2021_-_German_Language_2022-01-01.txt
```
These changes will not produce the expected result, since we definitely do want to use the date encoded in the file name if there is one. The issue to check here is if `rjson.packageHeader.fileNameDate` is the right place to look for the correct date which should be supplied somewhere in the POST request body.

PR #484, opened 2022-02-14
Fix ReviewsController.getEscalationTargetGroupId()

philboeselager:


server/gokbg3/grails-app/controllers/org/gokb/rest/ReviewsController.groovy
```diff
@@ -557,7 +557,9 @@ class ReviewsController {
         grailsApplication.config.gokb.centralGroups[componentClass] ?
         CuratoryGroup.findByNameIlike(grailsApplication.config.gokb.centralGroups[componentClass]) : null
     CuratoryGroup escalatedToCG = escalatingGroup.superordinatedGroup
-    if (!escalatedToCG && componentClass == "JournalInstance" && escalatingGroup != editorialGroup){
+    if (!escalatedToCG && escalatingGroup != editorialGroup &&
+        componentClass in [BookInstance.class.getSimpleName(), DatabaseInstance.class.getSimpleName(),
+                           JournalInstance.class.getSimpleName(), Package.class.getSimpleName()]){
```
hornmo: It would be better I think if we instead just check if there is an `editorialGroup` at all for this type of component (which would be set via the config).

Also in general the handling should be adjusted in such a way that it is clear that a review with no `escalatedToCG` is not an issue/error but an expected result which has to be conveyed as such.

hornmo (changes requested):

hornmo (approved):

PR #490, opened 2022-03-02
add routine for flagging invalid component names to housekeeping

hornmo:


server/gokbg3/grails-app/services/org/gokb/CleanupService.groovy
```diff
@@ -737,4 +741,44 @@ class CleanupService {
...
+    log.debug("Checking for corrupted component names")
+    TitleInstance.withNewSession { tsession ->
+      boolean more = true
+      int offset = 0
+      RefdataValue rr_type = RefdataCategory.lookup("ReviewRequest.StdDesc", "Invalid Name")
+      RefdataValue status_open = RefdataCategory.lookup("ReviewRequest.Status", "Open")
+      RefdataValue deleted_status = RefdataCategory.lookup('KBComponent.Status', KBComponent.STATUS_DELETED)
+
+      while (more) {
+        def batch = KBComponent.executeQuery("from KBComponent as kbc where name like '%�%' and status != :del and not exists (select 1 from ReviewRequest where componentToReview = kbc and stdDesc = :type and status = :status)", [type: rr_type, status: status_open, del: deleted_status], [max: 50])
```
philboeselager, line 755: @hornmo What is that? : 
'%�%'

hornmo: We are looking for the UTF-8 replacement character (U+FFFD), which may be produced for example when a KBART file that is not UTF-8 is opened as UTF-8 in Excel/Calc and then saved. We should also try to detect this on import (already added to UI), but this is a way to mark existing records retroactively.

philboeselager (approved):

PR #500, opened 2022-04-20
151 Elasticsearch 7 upgrade

philboeselager:


server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
```diff
@@ -669,86 +671,51 @@ class ESSearchService{
       }
 
       if( !errors && exactQuery.hasClauses() ) {
-        Client esclient = ESWrapperService.getClient()
```
hornmo, line 672: We still need to instantiate the client in this context.

server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
```diff
@@ -669,86 +671,51 @@ class ESSearchService{
...
-        SearchRequestBuilder es_request =  esclient.prepareSearch("exact")
 
-        es_request.setIndices(grailsApplication.config.gokb.es.indices.values() as String[])
-        es_request.setTypes(grailsApplication.config.globalSearch.types)
-        es_request.setQuery(exactQuery)
+        SearchRequest searchRequest = new SearchRequest(grailsApplication.config.searchApi.indices.values() as String[])
```
hornmo: The correct config value here should be `grailsApplication.config.gokb.es.indices.values()` instead.

hornmo (changes requested):
Generally there is no need to close the client after each interaction. Better to do this only on application shutdown to avoid unnecessary requests on the cluster.

PR #501, opened 2022-04-26
151 elasticsearch upgrade 7.15.1

philboeselager:


server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
```diff
@@ -669,86 +659,51 @@ class ESSearchService{
...
-
-            es_request.addSort(sortQry)
-          }
-        }
+        setSort(params, errors, searchSourceBuilder)
 
         if (!errors) {
-          search_action = es_request.execute()
+          searchRequest.source(searchSourceBuilder)
+          searchResponse = esClient.search(searchRequest, RequestOptions.DEFAULT)
```
hornmo: esClient is not defined.

hornmo (changes requested):

server/gokbg3/grails-app/services/com/k_int/ESSearchService.groovy
philboeselager: Sorry. Fixed.

hornmo (changes requested):
Looks fine functionally, but we should go for ES 7.10.2 for now.

hornmo (approved):

PR #507, opened 2022-05-18
268 search logic

philboeselager:
I think this fully addresses GOKbIT-268.
Not sure about the requirement "`OR` should be default operator for URLs", which I proposed myself. By now, we don't have any search for URLs, do we @hornmo ?
So this might be ready.


hornmo (changes requested):
I think it would be clearer just using the inbuilt functionality for changing the default operator here, see https://artifacts.elastic.co/javadoc/org/elasticsearch/elasticsearch/7.10.2/org/elasticsearch/index/query/QueryStringQueryBuilder.html#defaultOperator(org.elasticsearch.index.query.Operator)

philboeselager:
True. Replaced by https://github.com/openlibraryenvironment/gokb/pull/508 .

PR #508, opened 2022-05-23
Set ES queryStringQuery with AND logic (GOKbIT-268)

philboeselager:
... using the default operator.
Replacing https://github.com/openlibraryenvironment/gokb/pull/507 .


hornmo (commented):
This looks nice, we probably should do this for the other matchQueries over `name` & `altname` as well (in `processNameFields` as well as `processGenericFields`).

hornmo (approved):

PR #516, opened 2022-08-02
Replace colon by whitespace in Opensearch name fields (hotfix)

philboeselager:


server/gokbg3/grails-app/conf/elasticsearch/es_settings.json
```diff
@@ -18,6 +25,13 @@
             "autocomplete_filter"
           ]
         },
+        "custom_chars_analyzer": {
+          "type":      "custom",
+          "tokenizer": "standard",
+          "char_filter": [
+            "colon_filter"
+          ]
+        },
```
hornmo: To extend the standard analyzer we would also need `"filter": ["lowercase"]` here as per https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html

hornmo (changes requested):

hornmo (approved):

PR #517, opened 2022-08-09
fix suggest qfield & oai uri

hornmo:


philboeselager (approved):
We should probably make the port `80` configurable (one day).

PR #520, opened 2022-08-23
255 bulk import

hornmo:
Adds new job type for retrieving open EZB collections:
- API base from issue 255 has to be set as `gokb.ezbOpenCollections.url`
- For automated runs, set `gokb.ezbOpenCollections.enabled = true` (those are run after the scheduled package updates)
- For manual execution, use endpoint `admin/fetchEzbCollections`


philboeselager (approved):
Huge!

PR #521, opened 2022-09-28
Ezb check filesize

hornmo:


server/gokbg3/grails-app/controllers/org/gokb/rest/JobsController.groovy
```diff
@@ -223,7 +223,7 @@ class JobsController {
               startTime  : v.startTime,
               linkedItem : v.linkedItem,
               endTime    : v.endTime,
-              cancelled  : v.isCancelled()
+              cancelled  : v?.isCancelled()
```
philboeselager, line 226: We should elvisize the v.attribute calls as well, shouldn't we?

hornmo: Actually, this check isn't necessary here any more, since I added a check to `ConcurrencyManagerService.isCancelled()` directly. It is now: 

```
public boolean isCancelled() {
      return (this.task ? this.task.isCancelled() : true)
```

philboeselager (approved):

PR #523, opened 2022-10-24
254 validation endpoint

hornmo:


server/gokbg3/grails-app/services/org/gokb/ValidationService.groovy
```diff
@@ -0,0 +1,610 @@
...
+            message: "Skipped short/empty row!",
+            messageCode: "kbart.errors.shortRow"
+          ]
+        }
+
+        nl = csv.readNext()
+      }
+    }
+    else {
+      log.debug("Missing mandatory columns... skipping file processing!")
```
philboeselager, line 277: We should send a message here, no?

server/gokbg3/grails-app/services/org/gokb/ValidationService.groovy
```diff
@@ -0,0 +1,610 @@
...
+          result.rows.total++
+          result.rows.error++
+          result.errors.rows["${rowCount}"] = [
+            columnsCount: [
+              message: "Inconsistent column count (${nl.size()} <> ${header.size()})!",
+              messageCode: "kbart.errors.tabsCountRow",
+              args: [nl.size(), header.size()]
+            ]
+          ]
+          if (!result.errors.type['columnCount']) {
```
philboeselager: This little increase snippet (6 lines) occurs 3 times and is a good mini sub method (something like `increaseCount(def field)`) to make this method shorter, increase the code maintainabilty etc. We might need even more often in the future.

server/gokbg3/grails-app/services/org/gokb/ValidationService.groovy
```diff
@@ -0,0 +1,610 @@
...
+import org.apache.commons.validator.routines.UrlValidator
+
+import org.gokb.cred.*
+import org.gokb.GOKbTextUtils
+
+class ValidationService {
+
+  def TSVIngestionService
+
+  static final Map knownColumns = [
```
philboeselager: Static variables UPPER_CASED?

server/gokbg3/grails-app/services/org/gokb/ValidationService.groovy
```diff
@@ -0,0 +1,610 @@
...
+          ]
+          if (!result.errors.type['columnCount']) {
+            result.errors.type['columnCount'] = 1
+          }
+          else {
+            result.errors.type['columnCount']++
+          }
+          result.valid = false
+        }
+        else if (nl.size() >= 5) {
```
philboeselager: Instead of `5` we could add a static variable like `NUM_MAND_COLUMNS` which is statically initialized after the initialization of `knownColumns`.

server/gokbg3/grails-app/services/org/gokb/ValidationService.groovy
```diff
@@ -0,0 +1,610 @@
...
+      log.debug("Missing mandatory columns... skipping file processing!")
+    }
+
+    csv.close()
+
+    result
+  }
+
+  private CSVReader initReader (the_data) {
+    final CSVParser parser = new CSVParserBuilder()
```
philboeselager, line 296: A byte order mark (BOM) would break Ygor's processing, unless we had included BOM filtering. I guess we'd need that here as well?

server/gokbg3/grails-app/services/org/gokb/ValidationService.groovy
```diff
@@ -0,0 +1,610 @@
...
+    if (titleIdNamespace.value in ['isbn', 'pisbn']) {
+      try {
+        def valid_isbn = ISBN.parseIsbn(value)
+
+        result = value
+      }
+      catch(ISBNException ie) {}
+    }
+    else if (titleIdNamespace.value in ['issn', 'eissn']) {
+      def valid_issn = new ISSNValidator().isValid(value.toUpperCase())
```
philboeselager: Having a static ISSN_VALIDATOR would save some processing costs.

philboeselager (commented):
Looks very good. Some minor remarks.

philboeselager (commented):
Thumbs up.

philboeselager (approved):

PR #533, opened 2022-12-14
Add params.pkg filter to ComponentLookupService (GOKbIT-344)

philboeselager:


hornmo:
With the adjustments to https://github.com/openlibraryenvironment/gokb-ui/pull/104 this is not needed any more.

PR #554, opened 2023-05-31
Adjust import session handling

hornmo:
- Add bulk config search result layout
- Use service methods for public template kbart exports
- Fix variable in identifier update
- Fix EZB collection context
- Use temp files in KBART caching
- Adjust tests


philboeselager (approved):
Huge.
Code style maintenance - yeah!
The only thing I would change is placing the date (regular) expressions in constants. (PackageSourceUpdateService: 84, 135, 143, 150)

PR #559, opened 2023-08-02
512 insi

philboeselager:


server/grails-app/init/org/gokb/BootStrap.groovy
```diff
@@ -308,6 +308,13 @@ class BootStrap {
                     targetType: targetTypeJournal,
                     pattern: "^\\d+\$",
                     baseUrl: "https://ezb.uni-regensburg.de/detail.phtml?jour_id="
+                ],
+                [
+                    value: 'isni',
+                    name: 'ISNI',
+                    targetType: targetTypeJournal,
```
hornmo, line 315: The `targetTypeJournal` seems wrong, since the ISNI identifies Persons/Organisations. Should be `targetTypeOrg` instead.

hornmo (changes requested):

server/grails-app/init/org/gokb/BootStrap.groovy
philboeselager: Revisiting the code, I find that there is an "ISNI targetTypeOrg" configuration already existent in https://github.com/openlibraryenvironment/gokb/pull/559/files/3d45a86888a652e1da2e4d6ce9e7cc6e2cf667cf#diff-bc908ef684a866cf2668f7f76a11adb8c89da40b20239bde2032a6887f1540e1L278 . I guess we can close this PR unmerged. I'm not sure, where the ticket emerged from.

philboeselager: But no, then again, the base url that is mentioned in the new code, refers to journals. Not?

PR #561, opened 2023-08-08
Deproxy fixes

hornmo:


philboeselager (approved):

philboeselager:
The `withNewSession` expressions have been the key, right?

hornmo:
Thats part of it, but another one is the explicit use of `ClassUtils.deproxy(obj)`, especially in https://github.com/openlibraryenvironment/gokb/blob/9a323bb1328fb077f10f5b1a4f27988d3acaf3f3/server/src/main/groovy/org/gokb/DomainClassExtender.groovy#L328C13-L328C63

PR #626, opened 2025-06-02
Bug matching after wekb import

christophkohl1:
Wrong Identifiers are ignored on the Content Type given
Huge Packages with more than 30.000 Titles must be accepted by Admin before import
Conflicted Session Management for Title Matching Job was repaired

### DIFF
--- a/server/grails-app/services/org/gokb/PackageSourceUpdateService.groovy
+++ b/server/grails-app/services/org/gokb/PackageSourceUpdateService.groovy
@@ -85,7 +85,7 @@ class PackageSourceUpdateService {
 
       isExternalSourceImportOrUpdate = (pkg_source?.importConfig?.value && pkg_source.importConfig.value != "EZB")
       if ( isExternalSourceImportOrUpdate ) {
-        result.report = wekbIngestionService.startTitleImport(pkgInfo, pkg_source, pkg_plt, pkg_prov, p, job)
+        result.report = wekbIngestionService.startTitleImport(pkgInfo, pkg_source, pkg_plt, pkg_prov, p, job, async)
 
       } else {
         if (pkg_source?.url) {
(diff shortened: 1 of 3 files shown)


server/grails-app/services/org/gokb/PackageSourceUpdateService.groovy
```diff
@@ -79,13 +79,18 @@ class PackageSourceUpdateService {
       title_ns_serial_id = pkg_source?.titleIdSerial?.id ?: null
       title_ns_mono_id = pkg_source?.titleIdMonograph?.id ?: null
 
+      if ( restrictSize ) {
+        def ignoreSizeLimit = pkg_source.getIgnoreSizeLimit()
```
hornmo: Not sure how it would happen for there to be no source, but for safeties sake we should do `pkg_source?.ignoreSizeLimit`here.

server/grails-app/services/org/gokb/PackageSourceUpdateService.groovy
```diff
@@ -79,13 +79,18 @@ class PackageSourceUpdateService {
...
       if (job && !job.startTime) {
         job.startTime = startTime
       }
 
       isExternalSourceImportOrUpdate = (pkg_source?.importConfig?.value && pkg_source.importConfig.value != "EZB")
       if ( isExternalSourceImportOrUpdate ) {
-        result.report = wekbIngestionService.startTitleImport(pkgInfo, pkg_source, pkg_plt, pkg_prov, p, job)
+        result.report = wekbIngestionService.startTitleImport(pkgInfo, pkg_source, pkg_plt, pkg_prov, p, job, async, restrictSize)
 
       } else {
```
hornmo: Wouldn't it be clearer to just skip the whole variable thing and do `if (pkg_source?.importConfig?.value == 'WEKB')` here instead?

PR #684, opened 2026-02-04
688 ftp update

christophkohl1:


server/grails-app/controllers/org/gokb/rest/WebEndpointController.groovy
hornmo, line 1: All of the methods in this controller should be restricted to `ROLE_CONTRIBUTOR`.

server/grails-app/controllers/org/gokb/rest/WebEndpointController.groovy
```diff
@@ -0,0 +1,168 @@
...
+    def index() {
+        def result = [:]
+        def base = grailsApplication.config.getProperty('grails.serverURL') + "/rest"
+        User user = null
+
+        if (springSecurityService.isLoggedIn()) {
+            user = User.get(springSecurityService.principal?.id)
+        }
+
+        params['_embed'] = params['_embed'] ?: 'identifiedComponents'
```
hornmo: Not really applicable for this class of objects. Don't think we would have any embeds.

server/src/integration-test/groovy/org/gokb/UpdatePackageRunFTPSpec.groovy
```diff
@@ -0,0 +1,269 @@
...
+
+  def setup() {
+    ftpServer = new FakeFtpServer()
+    ftpServer.setServerControlPort(MOCKSERVER_PORT)
+    FileSystem fileSystem = new UnixFakeFileSystem()
+    fileSystem.add(new FileEntry(NON_KBART_FILE, CONTENTS))
+    ftpServer.setFileSystem(fileSystem)
+    ftpServer.addUserAccount(new UserAccount(USER, PASSWORD, HOME_DIR))
+
+    ftpServer.start()
```
hornmo: It might be best do do this once for all tests, and then make sure to only do the `stop()` after the last test.

server/grails-app/domain/org/gokb/cred/WebHookEndpoint.groovy
```diff
@@ -7,10 +7,14 @@ import groovy.util.logging.*
...
-  Long authmethod
-  String principal
-  String credentials
+  Long authmethod //legacy
+  RefdataValue supplyMethod //legacy
+  // RefdataValue transferMethod //rausnehmen - nur in Source
+  String principal //legacy
+  String credentials //legacy
   User owner
+  String ba_username
```
hornmo: Properties should be in camel case, which should then be explicitly mapped to a pascal case column name via `mapping`, for example `ep_url` with the prefix for endPoint . The prefix `ba` is too cryptic here and should be replaced by a more descriptive one.

server/grails-app/controllers/org/gokb/rest/WebEndpointController.groovy
```diff
@@ -0,0 +1,168 @@
...
+
+        result = componentLookupService.restLookup(user, WebHookEndpoint, params)
+
+        if (result.data) {
+            def resultList = result.data
+            resultList*.remove('ba_password')
+            resultList*.remove('ba_username')
+
+            if (params['method']) {
```
hornmo: Filtering after the fact is not permissible, since it may invalidate the pagination info. Custom logic should be added in `componentLookupService.restLookup` if necessary.

server/grails-app/controllers/org/gokb/rest/WebEndpointController.groovy
```diff
@@ -0,0 +1,168 @@
...
+            user = User.get(springSecurityService.principal?.id)
+        }
+
+        params['_embed'] = params['_embed'] ?: 'identifiedComponents'
+
+        result = componentLookupService.restLookup(user, WebHookEndpoint, params)
+
+        if (result.data) {
+            def resultList = result.data
+            resultList*.remove('ba_password')
```
hornmo: Mapping configuration for this should be set in the domain class via `jsonMapping.ignore`.

server/grails-app/controllers/org/gokb/rest/WebEndpointController.groovy
```diff
@@ -0,0 +1,168 @@
...
+        def start_db = LocalDateTime.now()
+
+        params['_embed'] = params['_embed'] ?: 'identifiedComponents'
+
+        result = componentLookupService.restLookup(user, WebHookEndpoint, params)
+
+        def resultList = result.data
+
+        if(resultList.size() > 0){
+            resultList*.remove('ba_password')
```
hornmo: See above

server/src/integration-test/groovy/org/gokb/UpdatePackageRunFTPSpec.groovy
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  void "test Mock FTP Server runs"() {
+
+    FTPClient ftpClient = new FTPClient()
+    ftpClient.connect("localhost", ftpServer.getServerControlPort())
+    ftpClient.login(USER, PASSWORD);
+
+    expect: "Directory exists"
+    ftpClient.listDirectories().length > 0
```
hornmo: Fails due to empty directories list.

hornmo (changes requested):

server/grails-app/controllers/org/gokb/rest/WebEndpointController.groovy
hornmo: This is still an issue. All filtering has to be done during the lookup.

server/grails-app/services/org/gokb/WebEndpointService.groovy
hornmo, line 1: Some formatting needed:

- No more than one blank line in a row
- Spacing in control statements (`if (xyz) {...}`)

hornmo (changes requested):

server/grails-app/services/org/gokb/WebEndpointService.groovy
```diff
@@ -0,0 +1,108 @@
...
+        def query = 'select whe from WebHookEndpoint whe '
+        def countQuery = 'select count(whe.id) from WebHookEndpoint whe '
+        def whereClause = 'where whe.url like :method'
+        def queryParams = [:]
+
+        if (params['method']) {
+            query += whereClause
+            countQuery += whereClause
+            queryParams['method'] = params['method'].toString().toLowerCase() + '%'
+        }
```
hornmo, line 32: Better to handle known cases here directly and avoid creating query params from user input.

hornmo (changes requested):

hornmo (approved):

PR #689, opened 2026-02-20
809 better monthly autoupdate cycle

christophkohl1:


server/grails-app/domain/org/gokb/cred/Source.groovy
```diff
@@ -175,11 +177,28 @@ class Source extends KBComponent {
       result = true
     }
     else if (frequency) {
-      use(TimeCategory) {
-        if (lastRun + intervals.get(frequency.value).days < new Date()) {
-          result = true
-        }
+
+      LocalDate lastRunDate = LocalDate.ofInstant(lastRun.toInstant(), ZoneId.systemDefault())
+      LocalDate now = LocalDate.ofInstant(new Date().toInstant(), ZoneId.systemDefault())
```
hornmo: Should be the same as `LocalDate.now()`.

hornmo (approved):

PR #690, opened 2026-02-20
add alerts for external editorial requests

hornmo:


server/grails-app/i18n/messages_de.properties
christophkohl1, line 1: independent of these changes, in this file the key 'spring.security.ui.forgotPassword.submit' exists twice

christophkohl1 (commented):
PR is approved but 1 other problem in file  server/grails-app/i18n/messages_de.properties

server/grails-app/i18n/messages_de.properties
hornmo: Fixed

PR #735, opened 2026-09-17
996 index cleanup job

christophkohl1:


server/grails-app/services/org/gokb/FTIndexCleanupService.groovy
```diff
@@ -0,0 +1,189 @@
...
+                till = Date.from(LocalDateTime.now().atZone(ZoneId.systemDefault()).toInstant())
+            }
+            else {
+                till = Date.from(updatedTill.atZone(ZoneId.systemDefault()).toInstant())
+            }
+
+            log.info("Start Syncing Tipps that were updated between: ... " + from + " - " + till)
+
+
+            List<TitleInstancePackagePlatform> tipps = TitleInstancePackagePlatform.executeQuery("select tipp from TitleInstancePackagePlatform as tipp where ( (tipp.lastUpdated > :us OR tipp.dateCreated > :us) AND tipp.lastUpdated <= :ut AND tipp.dateCreated <= :ut) order by tipp.lastUpdated, tipp.id", [us: from, ut: till], [readonly: true])
```
hornmo: Loading all TIPPs into the session at once is risky, depending on the date range. Optimally, do this in batches corresponding to the bulk index size instead, and do a `count(*)` query beforehand for the total. 

After each batch, the session should be cleaned up via:
`sessionFactory.getCurrentSession().clean()`

server/grails-app/services/org/gokb/FTIndexCleanupService.groovy
```diff
@@ -0,0 +1,189 @@
...
+            log.debug("REINDEX " + tippsToReindex.size() + " TIPPS ")
+
+            if (!dryRun) {
+                // Reindex not-up-to-date Tipps
+                Map updateResult = ftUpdateService.updateSpecifiedTippBulk(tippsToReindex, job)
+                if (updateResult.result != "OK") {
+                    completed = false
+                }
+                numberNewIndexedTipps = updateResult.indexed
+            }
```
hornmo: When adjusting to batch query, this should be run per batch.

server/grails-app/services/org/gokb/FTUpdateService.groovy
```diff
@@ -506,73 +520,183 @@ class FTUpdateService {
...
 
-  def triggerUpdateForClass(cls, Job j = null) {
-    log.debug("triggerUpdateForClass")
-    def indexType = ESWrapperService.indicesPerType[cls.simpleName]
 
-    if (this."${indexType}Running" == false) {
-      this."${indexType}Running" = true
-      this."${indexType}Update"(j)
-      log.debug("FTUpdate done.")
+  public void executeUpdateJobForIndex(Job j = null, RefdataValue jobType, JobDataMap dataMap ) {
```
hornmo: `jobType` should be the first parameter since it is mandatory.  `dataMap` should default to `null`  as well.

hornmo (changes requested):

